PR #36, opened 2022-04-08
chore: Update TTL data type to `NonZeroU64` and add publish workflow

### LINKED ISSUE #23: document behaviors for zero ttl
The API docs should make it clear what the behavior will be if TTL is explicitly set to zero. Currently we allow passing an Option<u32> as the TTL. Users familiar with memcache behavior might suspect that `Some(0)` will be treated as no expiration. However, it appears that it is treated as immediate expiration. Might make sense to take a `Option<NonZeroU32>` instead?

poppoerika:
This enforces users to pass `NonZeroU64::new(u64).unwrap()` to set default TTL or overwrite default TTL.
If users pass 0 to initialize `NonZeroU64` then `None` will be returned and thrown an error.
This adds an extra layer to make sure users pass an integer other than 0 before they blindly pass 0 without any thoughts.

Also added publish workflow to publish our SDK to crates.io as we merge prs to `main`.

Closes #23

### DIFF
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -1,5 +1,6 @@
 #[cfg(test)]
 mod tests {
+    use std::num::NonZeroU64;
     use std::{env, time::Duration};
 
     use momento::response::error::MomentoError;
@@ -12,7 +13,9 @@ mod tests {
 
     async fn get_momento_instance() -> SimpleCacheClient {
         let auth_token = env::var("TEST_AUTH_TOKEN").expect("env var TEST_AUTH_TOKEN must be set");
-        return SimpleCacheClient::new(auth_token, 5).await.unwrap();
+        return SimpleCacheClient::new(auth_token, NonZeroU64::new(5).unwrap())
+            .await
+            .unwrap();
     }
 
     #[tokio::test]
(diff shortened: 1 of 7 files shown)


tests/integration_test.rs
```diff
@@ -202,7 +210,9 @@ mod tests {
             + &jwt_payload_bad_data_plane_base64.clone()
             + "."
             + &jwt_invalid_signature_base_64.clone();
-        let mut client = SimpleCacheClient::new(bad_data_plane_jwt, 5).await.unwrap();
+        let mut client = SimpleCacheClient::new(bad_data_plane_jwt, NonZeroU64::new(5).unwrap())
```
eaddingtonwhite, line 213: so there is no way user can just pass `5` any more they have to explicitly cast to `NonZeroU64`? I really like the compile time check just kinda anoying user needs to pass this everytime now makes our code little more wordy.

poppoerika: Yes, now users need to pass `NonZeroU64::new(5).unwrap()` instead of just `5`.
Although passing 0 to `NonZeroU64::new(0).unwrap()` does not give a compile time error, users should be able to understand what get returned if they pass 0 based on [this](https://doc.rust-lang.org/stable/std/num/struct.NonZeroU64.html#method.new).

eaddingtonwhite (approved):

PR #39, opened 2022-04-12
chore: Add `new_momento_cli` to use in Momento CLI and add `agent` metadata for SDK

### LINKED ISSUE #25: Add `agent` metadata to `HeaderInterceptor` when starting a call
https://github.com/momentohq/client-sdk-javascript/issues/52

Also add `new_momento_cli` to be called only Momento CLI to initialize a Momento client with an appropriate agent value without changing the SDK experience for regular users.

poppoerika:
- Added a new function called `new_momento_cli` which creates a Momento client with the right agent value (`moment-cli:<VERSION>`). This function is intended to only be used in Momento CLI. In this way, no need to disrupt users' SDK experience but add different agent values for the SDK and Momento CLI.
- Removed `CacheHeaderInterceptor` and `AuthHeaderInterceptor` and created `HeaderInterceptor`.
- Added `agent` metadata for the SDK to be sent and displayed on [Grafana](https://g-1f6c1a3e90.grafana-workspace.us-west-2.amazonaws.com/d/y3H-lAy7z/sdk-usage-seen).
 
image

Related to https://github.com/momentohq/momento-cli/issues/102

Closes #25


src/simple_cache_client.rs
```diff
@@ -103,7 +103,7 @@ impl SimpleCacheClientBuilder {
     }
 
     pub fn build(self) -> SimpleCacheClient {
-        let agent_value = format!("rust:{}", VERSION).to_owned();
+        let agent_value = format!("rust:{}", VERSION);
```
eaddingtonwhite, line 106: should we make this `rust-cli` maybe? just wondering if we want be able break out CLI use vs someone using rust sdk for real in a project?

poppoerika: I think we are already on the same page about this but I was thinking to use this new function [here](https://github.com/momentohq/client-sdk-rust/pull/39/files#diff-03c0f9db265c085414e8facf02f2c1b0cf681cb91221b1170970cf91886f0f41R212) in Momento CLI so that when we initialize a client in the CLI, [this](https://github.com/momentohq/client-sdk-rust/pull/39/files#diff-03c0f9db265c085414e8facf02f2c1b0cf681cb91221b1170970cf91886f0f41R217) `agent_value` gets used instead of `rust:<VERSION>`.

src/simple_cache_client.rs
```diff
@@ -154,16 +166,66 @@ impl SimpleCacheClient {
...
+        let control_client = SimpleCacheClient::build_control_client(
+            auth_token.clone(),
+            momento_endpoints.control_endpoint,
+            agent_value.clone(),
+        )
+        .await;
+        let data_client = SimpleCacheClient::build_data_client(
+            auth_token.clone(),
+            momento_endpoints.data_endpoint,
+            agent_value,
```
eaddingtonwhite, line 180: why don't we need `agent_value.clone()` here but do for `build_control_client`?

src/grpc/header_interceptor.rs
```diff
@@ -0,0 +1,35 @@
...
+        for (key, value) in self.header.iter() {
+            if *key == *AUTHORIZATION {
+                request.metadata_mut().insert(
+                    tonic::metadata::AsciiMetadataKey::from_static(AUTHORIZATION),
+                    tonic::metadata::AsciiMetadataValue::from_str(value).unwrap(),
+                );
+            }
+            if !ARE_ONLY_ONCE_HEADER_SENT.load(Ordering::Relaxed) && *key != *AUTHORIZATION {
+                request.metadata_mut().insert(
+                    tonic::metadata::AsciiMetadataKey::from_static(AGENT),
```
eaddingtonwhite, line 27: Should this be? 
`tonic::metadata::AsciiMetadataKey::from_static(*key)`

What if we add another header to `ARE_ONLY_ONCE_HEADER_SENT` other then agent down line?

eaddingtonwhite (commented):
Looking good 2 small questions.

src/simple_cache_client.rs
poppoerika: Rust has the concept called ownership. (more details [here](https://doc.rust-lang.org/book/ch04-01-what-is-ownership.html))
So the first time, I passed `auth_token` to `build_control_client`, the clone of `auth_token` gets passed, which I believe necessary so that I can pass the actual `auth_token` to `build_data_client`. 
Once `auth_token` gets passed a different scope, it's now "owned" by that scope so cannot pass it to anywhere else.
But cloning values takes up some memory space, so I think we want to avoid cloning as much as possible. But that's when `clippy` comes in handy. If it finds redundant use of clone, then it will be linted.

src/grpc/header_interceptor.rs
poppoerika: Since `tonic::metadata::AsciiMetadataKey::from_static` takes only a static variable, `AGENT` needs to get passed, I believe. (At least passing `*key` is not valid)
Also `AGENT` is a constant variable which lives the entire lifetime of a program, so it can be used later like the below example:
```
if !ARE_ONLY_ONCE_HEADER_SENT.load(Ordering::Relaxed) && *key != *AUTHORIZATION {
     request.metadata_mut().insert(
     tonic::metadata::AsciiMetadataKey::from_static(AGENT),
     tonic::metadata::AsciiMetadataValue::from_str(value).unwrap(),
      );
      ARE_ONLY_ONCE_HEADER_SENT.store(true, Ordering::Relaxed);
}
request.metadata_mut().insert(
     tonic::metadata::AsciiMetadataKey::from_static(AGENT),
     tonic::metadata::AsciiMetadataValue::from_str(value).unwrap(),
);
```

eaddingtonwhite: shouldn't you switch up logic in if statement then to be like
`if !ARE_ONLY_ONCE_HEADER_SENT.load(Ordering::Relaxed) && *key == *AGENT {`

poppoerika: Yup, it makes sense. 
There might be a better way to handle new metadata in the future but I spent some time but this `&mut self` thing is a bit confusing and needing to pass a static variable was the part I just time boxed yesterday.
I think it's okay for now to specifically call `AGENT` we can come back and update code structure later if needed.

eaddingtonwhite (approved):

PR #43, opened 2022-04-12
feat: allow changing builder default ttl

brayniac:
Allow the `SimpleCacheClientBuilder` default ttl to be modified by
adding a new function. This is necessary if we want to be able to
create several clients with different ttls after initializing the
builder.

For example, this is desirable in the proxy where we want to allow
a different ttl for each cache but use the same underlying grpc
channels.


brayniac:
Replaces #42

bruuuuuuuce (approved):
:shipit:

PR #48, opened 2022-04-22
chore: add Deserialize and Serialize to SigningKey for serde_json support

tylerburdsall:
Adds `Serialize` and `Deserialize` so that consumers can easily pretty-print results of a `ListSigningKeysResponse`


virratanasangpunth (approved):
I dunno much about serde serialization library but if it does the job this seems fine to me!

PR #51, opened 2022-05-17
feat: update client_protos, clean up result

kvcache:
* delete vendored .proto files
* delete generated grpc files
* delete generation function in build.rs
* update dependencies
  * momento-protos
  * tonic
* remove superfluous dependencies
  * prost
  * rustls
  * webpki
  * tower
* reclassify test dependencies as test dependencies
  * tokio
  * base64-url
* add gomomento homepage to package manifest
* fix code with updates
  * fix warning for deprecated from_str by using standard try_from
  * remove pointless HashMap from HeaderInterceptor, remove pointless
    consts in 2 files. Will reduce memory usage by a few bytes and
    speed up requests by about a half a nanosecond while making the
    code simpler and not coupled to 2 files' magic consts. Also
    deduplicates and simplifies interceptor map creation. I would
    not have included this change with the rest but it overlapped
    with other changes and this was the easiest resolution.
  * similarly, deduplicate connect_channel copypasta. Let's not
    copypaste so much okay friendos?

A lot of changes in this change; this is a cleanup PR, but it's not
a complete cleanup. I've only messed with the things that overlap
with the client_protos update but there's more to do in here.


tylerburdsall (approved):
Thanks for simplifying the header interceptor stuff!

PR #53, opened 2022-05-18
chore: clean up SimpleCacheClientBuilder

kvcache:
* delete a bunch of copy/paste code (let's slow down on that a little)
* add log dependency and inaugural 1 log message that helped me
* add env_logger dev-dependency so a developer can enable logging in tests if they want
* change SimpleCacheClientBuilder's new() to have the same semantics as our
  pre-existing copy/paste method family for creating SimpleCacheClient. It is now lazy
  and will allow endpoint resolution to only happen when an interaction is performed,
  not eagerly (will create a data-only client more quickly now)
* change rust user agents: rust-sdk:VERSION is now the default, and we now have
  rust-integration_test:VERSION and will soon have rust-cli:VERSION.
* fix up tests

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -9,16 +9,11 @@ use momento_protos::{
 use serde_json::Value;
 use std::convert::TryFrom;
 use std::num::NonZeroU64;
-use tonic::{
-    codegen::InterceptedService,
-    transport::{Channel, ClientTlsConfig, Uri},
-    Request,
-};
+use tonic::{codegen::InterceptedService, transport::Channel, Request};
 
-use crate::grpc::header_interceptor::HeaderInterceptor;
-use crate::{endpoint_resolver::MomentoEndpointsResolver, utils::connect_channel};
+use crate::endpoint_resolver::MomentoEndpointsResolver;
+use crate::{grpc::header_interceptor::HeaderInterceptor, utils::connect_channel_lazily};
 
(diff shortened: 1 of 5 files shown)


src/simple_cache_client.rs
```diff
@@ -61,44 +56,61 @@ pub struct SimpleCacheClientBuilder {
...
-            .unwrap()
-            .connect()
-            .await?;
+        let control_channel = connect_channel_lazily(&momento_endpoints.control_endpoint)?;
+        let data_channel = connect_channel_lazily(&momento_endpoints.data_endpoint)?;
 
         match utils::is_ttl_valid(&default_ttl_seconds) {
             Ok(_) => Ok(Self {
-                data_endpoint,
+                data_endpoint: momento_endpoints.data_endpoint,
```
wandaitzuchen, line 108: We include in the create_signing_key response the data endpoint as is from the auth token (i.e. not `momento_endpoints.data_endpoint`, which added `"https://{}:443"`). The signing key integ test should fail with this code change. 
To avoid parsing auth token twice we can update `MomentoEndpointsResolver::resolve` to take a parsed claim instead of the token?

kvcache: I've reworked MomentoEndpointsResolver and removed a bunch of copypasta. That should pass now, thanks for the note.

wandaitzuchen (approved):

PR #56, opened 2022-05-25
chore: Align MomentoCacheListResponse and list_caches

malandis:
The method `SimpleCacheClient::list_caches` supports pagination. It does this by means of
the parameter `next_token`: the client starts iteration with this set to
`None`, and ends when `MomentoCacheListResponse::next_token` is `None`.

Previously `MomentoCacheListResponse::next_token` was not optional, so
the end sentinel was empty string as opposed to `None`. This means the
client would have to start iteration with `None` and test for end
iteration comparing to empty string, which is confusing.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -241,7 +241,10 @@ impl SimpleCacheClient {
             .collect();
         let response = MomentoListCacheResult {
             caches,
-            next_token: res.next_token,
+            next_token: match res.next_token.is_empty() {
+                true => None,
+                false => Some(res.next_token),
+            },
         };
         Ok(response)
     }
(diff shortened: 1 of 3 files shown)


tylerburdsall (approved):
Looks great!

src/simple_cache_client.rs
```diff
@@ -241,7 +241,10 @@ impl SimpleCacheClient {
             .collect();
         let response = MomentoListCacheResult {
             caches,
-            next_token: res.next_token,
+            next_token: match res.next_token.is_empty() {
+                true => None,
+                false => Some(res.next_token),
+            },
```
kvcache, line 247: Can you explain what this change does?

kvcache: Chatted offline. This is fine.

kvcache (approved):

PR #59, opened 2022-06-03
feat(#49): Adds New Delete Function to SDK

### LINKED ISSUE #49: [New API] Add Delete API
Our backend server now supports delete, this functionality needs to be added into the SDK.

Need to update to use newest proto: https://github.com/momentohq/client_protos/blob/main/proto/cacheclient.proto#L35

WillRInternship:
Closes #49

This pull request adds new delete functionality to the rust sdk.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -63,17 +63,16 @@ pub fn request_meta_data<T>(
     request: &mut tonic::Request<T>,
     cache_name: &str,
 ) -> Result<(), MomentoError> {
-    Ok(tonic::metadata::AsciiMetadataValue::try_from(cache_name)
+    tonic::metadata::AsciiMetadataValue::try_from(cache_name)
         .map(|value| {
             request.metadata_mut().append("cache", value);
-            ()
         })
         .map_err(|e| {
             MomentoError::InvalidArgument(format!(
                 "Could not treat cache name as a header value: {}",
-                e.to_string()
+                e
             ))
-        })?)
+        })
 }
 
 impl SimpleCacheClientBuilder {
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -59,6 +59,13 @@ pub struct SimpleCacheClientBuilder {
     user_agent_name: String,
 }
 
+pub fn request_meta_data<T>(request: &mut tonic::Request<T>, cache_name: &str) {
+    request.metadata_mut().append(
+            "cache",
+            tonic::metadata::AsciiMetadataValue::try_from(cache_name).unwrap(),
+    );
+}
+
 impl SimpleCacheClientBuilder {
     /// Returns a builder which can produce an instance of a Momento client
     ///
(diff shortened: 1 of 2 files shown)


bruuuuuuuce (dismissed):
:shipit: nice work, looks good!

src/simple_cache_client.rs
```diff
@@ -459,4 +459,54 @@ impl SimpleCacheClient {
...
+        key: I,
+    ) -> Result<(), MomentoError> {
+        utils::is_cache_name_valid(cache_name)?;
+        let mut request = tonic::Request::new(DeleteRequest {
+            cache_key: key.into_bytes(),
+        });
+        request.metadata_mut().append(
+            "cache",
+            tonic::metadata::AsciiMetadataValue::try_from(cache_name).unwrap(),
+        );
```
malandis, line 501: We should refactor this into a separate function. Adding the cache metadata is done by two other functions in this module (`get` and `set`).

src/simple_cache_client.rs
```diff
@@ -59,6 +59,13 @@ pub struct SimpleCacheClientBuilder {
     user_agent_name: String,
 }
 
+pub fn request_meta_data<T>(request: &mut tonic::Request<T>, cache_name: &str) {
+    request.metadata_mut().append(
+        "cache",
+        tonic::metadata::AsciiMetadataValue::try_from(cache_name).unwrap(),
+    );
+}
```
kvcache, line 67: Unwrap is an antipattern in SDK's. It causes user applications to crash for things that, really, we have no idea whether the user should crash over.

Consider plumbing the MomentoError up like
```
pub fn request_meta_data<T>(request: &mut tonic::Request<T>, cache_name: &str) -> Result<(), MomentoError> {
    Ok(
        tonic::metadata::AsciiMetadataValue::try_from(cache_name)
            .map(|value| {
                request.metadata_mut().append("cache", value);
                ()
            })
            .map_err(|e| {
                MomentoError::InvalidArgument(format!("Could not treat cache name as a header value: {}", e.to_string()))
            })?
    )
}
```

And when you call request_meta_data you call it like
```
request_meta_data(&mut request, cache_name)?;
```

so that when a bad value is encountered you bail upward from there.

eaddingtonwhite (approved):
Nice this is LGTM! thanks for all refactor as well DRY up the code 👍 Would let @kvcache and @malandis take one more pass also but nice work 🚢

src/simple_cache_client.rs
```diff
@@ -391,10 +407,7 @@ impl SimpleCacheClient {
             cache_body: body.into_bytes(),
             ttl_milliseconds: ttl_to_use,
         });
-        request.metadata_mut().append(
-            "cache",
-            tonic::metadata::AsciiMetadataValue::try_from(cache_name).unwrap(),
-        );
+        request_meta_data(&mut request, cache_name)?;
```
kvcache, line 410: Thanks for getting rid of these unwraps

kvcache (approved):

malandis (approved):

bruuuuuuuce (approved):
:shipit:

PR #60, opened 2022-07-13
feat: Add support for momento endpoints

gautamomento:
Allow using specified momento endpoints instead of retrieving the endpoints from the authentication token.


src/jwt.rs
```diff
@@ -6,11 +6,11 @@ use crate::response::error::MomentoError;
...
 pub struct Claims {
     pub sub: String,
-    pub c: String,
-    pub cp: String,
+    pub c: Option<String>,
+    pub cp: Option<String>,
 }
 
-pub fn decode_jwt(jwt: &str) -> Result<Claims, MomentoError> {
+pub fn decode_jwt(jwt: &str, momento_endpoint: &Option<String>) -> Result<Claims, MomentoError> {
```
kvcache, line 13: Why are you passing a reference of Option instead of Option?

kvcache (commented):
looks good. Got a couple nits

src/simple_cache_client.rs
```diff
@@ -96,11 +96,28 @@ impl SimpleCacheClientBuilder {
...
             auth_token,
             default_ttl_seconds,
             "sdk",
+            None,
+        )
+    }
+
+    /// Like new() above, but requires a momento_endpoint.
+    // TODO: Update the documentation
+    pub fn new_with_endpoint(
```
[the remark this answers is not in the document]
gautamomento: I am going to hold off making the existing method deprecated till we publish the endpoints either in documentation or via the SDK.

src/endpoint_resolver.rs
```diff
@@ -83,10 +82,8 @@ impl MomentoEndpointsResolver {
...
     fn hosted_zone_endpoint(hosted_zone: Option<String>, prefix: &str) -> Option<MomentoEndpoint> {
-        if hosted_zone.is_none() {
-            return None;
-        }
-        let hostname = format!("{}{}", prefix, hosted_zone.clone().unwrap());
+        hosted_zone.as_ref()?;
+        let hostname = format!("{}{}", prefix, hosted_zone.unwrap());
         Some(MomentoEndpointsResolver::wrapped_endpoint(
             prefix, hostname, "",
         ))
```
kvcache, line 89: probably a more rusty way to option-coalesce would be something like 
```suggestion
fn hosted_zone_endpoint(hosted_zone: Option<String>, prefix: &str) -> Option<MomentoEndpoint> {
        hosted_zone.map(|hosted_zone| {
            let hostname = format!("{}{}", prefix, hosted_zone);
            MomentoEndpointsResolver::wrapped_endpoint(prefix, hostname, "")
        })
```

which avoids unwrap() and is shorter. The `.map` thing on Option is pretty nice https://doc.rust-lang.org/std/option/enum.Option.html#method.map

src/simple_cache_client.rs
```diff
@@ -96,11 +96,28 @@ impl SimpleCacheClientBuilder {
     ///         .build();
     /// # })
     /// ```
+    // TODO: Add test with token and endpoint
```
rlinehan, line 99: Is this something you want to add in this PR, or is this for later?

src/simple_cache_client.rs
```diff
@@ -96,11 +96,28 @@ impl SimpleCacheClientBuilder {
...
         SimpleCacheClientBuilder::new_with_explicit_agent_name(
             auth_token,
             default_ttl_seconds,
             "sdk",
+            None,
+        )
+    }
+
+    /// Like new() above, but requires a momento_endpoint.
+    // TODO: Update the documentation and deprecate the existing new method.
```
rlinehan, line 110: wonder if it would be worth adding a `once xyz happens` to this TODO so people reading this know when we plan to actually get to it?

rlinehan (approved):
🙈 left 2 comments about `TODO`s but they're just little nitpicks, not blocking.

rlinehan (approved):

src/endpoint_resolver.rs
```diff
@@ -58,33 +69,32 @@ impl MomentoEndpointsResolver {
         MomentoEndpointsResolver::wrapped_endpoint("https://", hostname, ":443")
     }
 
-    fn get_control_endpoint(claims: &Claims, hosted_zone: &Option<String>) -> MomentoEndpoint {
-        MomentoEndpointsResolver::get_control_endpoint_from_hosted_zone(hosted_zone)
-            .unwrap_or_else(|| MomentoEndpointsResolver::https_endpoint(claims.cp.to_owned()))
+    fn get_control_endpoint(claims: &Claims, hosted_zone: Option<String>) -> MomentoEndpoint {
+        MomentoEndpointsResolver::get_control_endpoint_from_hosted_zone(hosted_zone).unwrap_or_else(
+            || MomentoEndpointsResolver::https_endpoint(claims.cp.as_ref().unwrap().to_owned()),
```
kvcache, line 74: so if a user does not pass in a `--endpoint` and they're using a new token they'll get a PANIC here, right? This should be cleaned up in a follow-up pr, feel free to open an issue if you want to defer

kvcache (approved):
got some `.unwrap()` hygiene we're going to need to do in this sdk but it doesn't block this change.

src/endpoint_resolver.rs
gautamomento: This method will fail in isolation but in the overall scheme of things if endpoint isn't provided with new token, customers will see `"Could not parse token. Please ensure a valid token was entered correctly."`

When endpoint isn't provided, the JWT decode expects `cp` and `c` claims to be present in the token.

https://github.com/momentohq/client-sdk-rust/blob/main/src/jwt.rs#L25-L28

gautamomento: Nonetheless, this should be cleaned up.
https://github.com/momentohq/client-sdk-rust/issues/61

PR #68, opened 2022-11-17
chore: dep scanning workflow

eaddingtonwhite:


bruuuuuuuce (approved):
:shipit:

cprice404 (commented):
i'm on board with this but it is failing because it caught something

PR #73, opened 2022-12-08
chore: makes clippy check consistent for unwrap warning

eaddingtonwhite:


tylerburdsall (commented):
Should we also address some of these warnings Clippy found? We also should create a config file that allows us to use `.unwrap()` in tests, ping me offline and I can show you an example

tylerburdsall (approved):

PR #75, opened 2022-12-19
feat: add dictionary get and set commands

brayniac:
Adds dictionary get and set commands

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ homepage = "https://gomomento.com/"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-momento-protos = { version = "0.18.1" }
+momento-protos = { version = "0.37.0" }
 log = "0.4.17"
 tonic = { version = "0.7.2", features = ["tls", "tls-roots", "tls-webpki-roots"] }
 jsonwebtoken = "8.0.1"
--- a/README.md
+++ b/README.md
@@ -14,6 +14,15 @@ Rust SDK for Momento, a serverless cache that automatically scales without any o
 
 <br/>
 
+## Preview Features
+
+This SDK contains APIs for interacting with collection data structures: Lists, Sets, and Dictionaries.  These APIs
+are currently in preview.  If you would like to request early access to the data structure APIs, please contact us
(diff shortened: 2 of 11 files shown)


brayniac:
As far as I can tell, this isn't failing due to any code added. Looks like some clippy changes causing lint failures and a "malformed auth token" causing doctest failures.

tylerburdsall:
> As far as I can tell, this isn't failing due to any code added. Looks like some clippy changes causing lint failures and a "malformed auth token" causing doctest failures.

@brayniac That's right, we're also looking to merge https://github.com/momentohq/client-sdk-rust/pull/73 after some changes are made so that clippy warnings are more in-line with our code standards.

Thank you for adding these changes! We can ensure that we add a disclaimer for the `dictionary_*` APIs since they are currently under Preview and require contacting support to enable those APIs for a cache. If you're okay making the suggested Clippy changes we can merge your PR

brayniac:
Thanks @tylerburdsall - working through the clippy lints. Any idea what we need to do to get doctests to pass? Things looked good locally with my own auth token, with exception of the signing key limit eventually causing issues for me with repeated test runs.

tylerburdsall:
@brayniac we'd need to allow-list your auth token, although we can always check out your changes and confirm they work once the Clippy changes are addressed.

We (currently) use this command to check for clippy warnings, so you can always use that to speed up local development:

```bash
cargo clippy --all-targets --all-features -- -D warnings -W clippy::unwrap_used
```

Oh, and something I was looking to add that might speed up the work was to add a `clippy.toml` file in the root of the project with:

```toml
allow-expect-in-tests = true
```

so that you don't have to worry about modifying a bunch of tests.

If you're interested in verifying the workflow yourself, feel free to reach out to @eaddingtonwhite or email `support@momentohq.com` and we can get you set up 😁

brayniac:
@tylerburdsall - the doctests fail in GitHub workflow because of my auth token?

tylerburdsall:
@brayniac Ah apologies, I got ahead of myself. I do see that they are failing because of a malformed auth token, but I would suspect that if that got resolved and you had a non allow-listed token then the tests would fail.

If the build is failing only because of a malformed auth token I can help debug that

brayniac:
Closing in favor of #76

Issue #80, opened 2023-01-06
investigate idle connection behavior after NLB timeout


cprice404:
We recently had an issue with the javascript SDK where, if a client object was kept open but idle for > the NLB idle connection timeout (350s), then it would end up in a bad state and the next request would fail.

Need to see if we can reproduce this and then see if keepalive settings will fix it.

cprice404:
I'm able to reliably reproduce a transport error on the first request after the idle timeout with this code.  Going to experiment with keepalive settings.

```
use momento::response::cache_get_response::MomentoGetStatus;
use momento::simple_cache_client::{SimpleCacheClient, SimpleCacheClientBuilder};
use std::{env, thread, time};
use std::num::NonZeroU64;
use std::process;
use std::str;
use std::str::Utf8Error;

async fn do_get_set(cache_client: &mut SimpleCacheClient) {
    let cache_name = "tacocache";

    // Sets key with default TTL and get value with that key
    let key = String::from("my_key");
    let value = String::from("my_value");

    match cache_client
        .set(&cache_name, key.clone(), value.clone(), None)
        .await
    {
        Ok(_) => {}
        Err(err) => {
            eprintln!("{}", err);
        }
    };

    match cache_client.get(&cache_name, key.clone()).await {
        Ok(r) => match r.result {
            MomentoGetStatus::HIT => match str::from_utf8(&r.value) {
                Ok(s) => println!("cache hit! result: {}", s),
                Err(_) => println!("Error converting result to utf-8 string!")
            }
            MomentoGetStatus::MISS => println!("cache miss"),
            _ => println!("error occurred"),
        },
        Err(err) => {
            eprintln!("{}", err);
        }
    };

}

#[tokio::main]
async fn main() {
    // Initializing Momento
    let auth_token =
        env::var("MOMENTO_AUTH_TOKEN").expect("env var MOMENTO_AUTH_TOKEN must be set");
    let item_default_ttl_seconds = 60;
    let mut cache_client = match SimpleCacheClientBuilder::new(
        auth_token,
        NonZeroU64::new(item_default_ttl_seconds).expect("expected a non-zero number"),
    ) {
        Ok(client) => client,
        Err(err) => {
            eprintln!("{}", err);
            process::exit(1);
        }
    }
    .build();

    do_get_set(&mut cache_client).await;
    println!("Sleeping for 6 minutes to try to trigger NLB idle timeout");
    thread::sleep(time::Duration::from_secs(7 * 60));
[...]

PR #81, opened 2023-01-07
fix: add keepalive settings to preserve idle connections

cprice404:
Prior to this commit, we did not have keepalives enabled for
idle gRPC channels.  As a result, any channel that was idle
for longer than NLB's idle connection timeout (about 6 minutes)
would be closed by the server, and the next attempt to use
it in the client would result in a transport error.

This commit enables keepalive for idle connections, sending
keepalive pings once every 2 minutes.  In local testing I observed
that a channel I left idle for over 6 minutes was still able
to be used successful on the next request after the idle period.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -10,10 +10,8 @@ _Read this in other languages_: [日本語](README.ja.md)
 - A Momento Auth Token is required, you can generate one using the [Momento CLI](https://github.com/momentohq/momento-cli)
 
 ```bash
-cargo build
-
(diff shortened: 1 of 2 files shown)


cprice404:
I have tested this locally and verified that it fixes the issue that was demonstrated by the reproducer code from https://github.com/momentohq/client-sdk-rust/issues/80 .

eaddingtonwhite (approved):
Thanks @cprice404

PR #86, opened 2023-01-23
fix: remove import for non-existant path in doctest

swlynch99:
`cargo test` is failing because the doctests try and use a path that doesn't exist. Nothing in the doctest is actually using the imported item so I've just removed the import. This also fixes a type inference error in the same test.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -782,7 +782,6 @@ impl SimpleCacheClient {
     /// # tokio_test::block_on(async {
     ///     use std::env;
     ///     use momento::{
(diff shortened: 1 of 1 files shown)


src/simple_cache_client.rs
```diff
@@ -806,7 +805,7 @@ impl SimpleCacheClient {
     ///     let resp = momento.dictionary_delete(
     ///         &cache_name,
     ///         &dictionary_name,
-    ///         Fields::All,
+    ///         Fields::<Vec<u8>>::All,
```
kvcache, line 808: It always feels weird to me in rust to see an enum specialized by a turbofish.

kvcache (approved):

PR #88, opened 2023-01-24
feat: allow dictionary client methods to take non-utf-8 dictionary names

swlynch99:
As of now, the dictionary methods on `SimpleCacheClient` take the dictionary name as a `&str` and so they require that the dictionary name is valid UTF-8. As I understand it, momento has no such restriction and the protocol appears to allow any sequence of bytes as a name. This PR changes the dictionary argument to take anything that implements `MomentoRequest` and expands what implements `MomentoRequest` enough to cover all previous calling patterns.

This change came out of some comments on https://github.com/pelikan-io/pelikan/pull/12

This is a breaking change (although most users should be fine) so I have marked it as a feature.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -33,43 +33,16 @@ use crate::utils;
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
-pub trait MomentoRequest {
+pub trait IntoBytes {
     fn into_bytes(self) -> Vec<u8>;
 }
 
-impl MomentoRequest for String {
+impl<T> IntoBytes for T
+where
+    T: Into<Vec<u8>>,
+{
     fn into_bytes(self) -> Vec<u8> {
-        self.into_bytes()
-    }
-}
-
-impl MomentoRequest for Vec<u8> {
-    fn into_bytes(self) -> Vec<u8> {
-        self
-    }
-}
-
-impl MomentoRequest for &str {
-    fn into_bytes(self) -> Vec<u8> {
-        self.to_string().into_bytes()
-    }
-}
-
-impl MomentoRequest for &[u8] {
-    fn into_bytes(self) -> Vec<u8> {
-        self.to_owned()
-    }
-}
-
-impl MomentoRequest for &Vec<u8> {
-    fn into_bytes(self) -> Vec<u8> {
-        self.clone().into_bytes()
-    }
-}
-
-impl MomentoRequest for &String {
-    fn into_bytes(self) -> Vec<u8> {
-        self.clone().into_bytes()
+        self.into()
     }
 }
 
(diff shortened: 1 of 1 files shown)


brayniac (dismissed):
For what it's worth, this looks good to me.

src/simple_cache_client.rs
```diff
@@ -55,6 +55,24 @@ impl MomentoRequest for &str {
...
+    fn into_bytes(self) -> Vec<u8> {
+        self.clone().into_bytes()
+    }
+}
+
+impl MomentoRequest for &String {
+    fn into_bytes(self) -> Vec<u8> {
+        self.clone().into_bytes()
+    }
+}
```
kvcache, line 74: Instead of defining "Into"-like traits that hide expensive payload copies from & to standard types, should we just redefined our trait to use the standard definitions for type conversions?
```rust
pub trait MomentoRequest : Into<Vec<u8>> {}
impl <T> MomentoRequest for T where T : Into<Vec<u8>> {} // I'm not sure if the blanket impl is actually needed on this one
```

That would let you delete like 20 lines =) and would make it more clear the intent that this is a from/into thing. If a user wants to clone before passing in to us that's always an option, but in Rust we should be clear about what we need at the call site - which is a wholly owned `Vec<u8>` or something that can be converted `Into` it.

kvcache: Also, the way this is used, both existing and in your new change, is not a "request" it's a "MomentoBytes" or just "momento::Bytes." It needs to be renamed...

src/simple_cache_client.rs
```diff
@@ -543,10 +561,10 @@ impl SimpleCacheClient {
     ///     momento.delete_cache(&cache_name).await;
     /// # })
     /// ```
-    pub async fn dictionary_set<K: MomentoRequest, V: MomentoRequest>(
+    pub async fn dictionary_set<K: MomentoRequest, V: MomentoRequest, D: MomentoRequest>(
         &mut self,
         cache_name: &str,
-        dictionary_name: &str,
+        dictionary_name: D,
```
kvcache, line 567: nit: I don't know if I'm the only one, but these 1char generics on a parameter get me a little. `dictionary_name: impl MomentoRequest,` would read more nicely imo and would make the method signature more friendly.

kvcache (commented):
I'm not so sure about the trait you're reusing here. I think it needs to be updated before it's really suitable to be "a cache key for referencing a dictionary"

src/simple_cache_client.rs
swlynch99: I'm fine with this. The `MomentoRequest` trait has been a misnomer since I started looking at the code, uh, yesterday. I was just trying to follow along with the existing code.

I'll update the trait to be called `IntoBytes` and change it's uses to suit.

src/simple_cache_client.rs
swlynch99: I've updated this for the new parameters added in this PR. If you're open to it, I'll make a second PR to replace existing uses where appropriate.

kvcache: > If you're open to it, I'll make a second PR to replace existing uses where appropriate.

Of course, and with my thanks! ❤️

kvcache (approved):
Looks good, thanks for the improvement

PR #90, opened 2023-01-24
feat: Flatten `momento::response` module into a flat public API

swlynch99:
The response module has a whole bunch of submodules that don't really add much information and make things much harder to find and use. This PR flattens the public API down into just the `response` module while still keeping the other modules for code organization internally.


src/simple_cache_client.rs
```diff
@@ -17,17 +17,11 @@ use crate::endpoint_resolver::MomentoEndpointsResolver;
...
-    create_signing_key_response::MomentoCreateSigningKeyResponse,
-    error::MomentoError,
-    list_cache_response::{MomentoCache, MomentoListCacheResult},
-    list_signing_keys_response::{MomentoListSigningKeyResult, MomentoSigningKey},
+    MomentoCache, MomentoCreateSigningKeyResponse, MomentoDictionaryFetchResponse,
+    MomentoDictionaryFetchStatus, MomentoDictionaryGetResponse, MomentoDictionaryGetStatus,
+    MomentoDictionarySetResponse, MomentoDictionarySetStatus, MomentoError, MomentoGetResponse,
+    MomentoGetStatus, MomentoListCacheResult, MomentoListSigningKeyResult, MomentoSetResponse,
+    MomentoSetStatus, MomentoSigningKey,
 };
```
swlynch99, line 25: If it was up to me I'd just convert this to a glob import but I'm not entirely sure if we're ok with doing that in this codebase.

kvcache: > If it was up to me I'd just convert this to a glob import but I'm not entirely sure if we're ok with doing that in this codebase.

We don't really do glob imports. It's a small character count savings trading for symbol location ambiguity. It is sad for people who are reading without an IDE.

swlynch99: 👍

That's why I checked :)

swlynch99:
Fixed the compile errors but I had to bring the example in under a workspace so that it was actually using the current version of the momento crate. If you'd like, I can undo this but putting it under a workspace will help make sure the example stays up to date.

kvcache (dismissed):
I like this and I particularly like the examples being bound to the repo instead of being allowed to diverge. @cprice404 what do you think?

swlynch99:
Another option to the workspace is to make use of cargo's example directory (see how the [io-uring](https://github.com/tokio-rs/io-uring) crate does it).

cprice404:
> I like this and I particularly like the examples being bound to the repo instead of being allowed to diverge. @cprice404 what do you think?

For now, for better or for worse, we are using the examples directory in conjunction with dependabot to make sure that when official releases are published to crates.io, we have some automation that is ensuring that the latest package is usable from there rather than from local source.

Until we come up with a better solution for this I would prefer that we keep the Momento dependency version in the examples/Cargo.toml pinned to an explicit version.  I recognize that for local development you may need to flip it in order to make things easy to iterate on but for now I think we need to keep the version that is committed to git pinned to a specific version.

malandis:
Good change flattening `momento::response`. This is what we have done in in the other languages (.NET, JavaScript, Python, ...) to reduce import clutter.

swlynch99:
I've gone and removed the changes to the examples directory. I'll make a follow-up PR to update the examples directory to the latest version once a new version is released.

Also, I'm not sure what's up with CI because the clippy errors I'm getting don't make much sense...

swlynch99:
I've gone and renamed `examples -> example` so that running `cargo clippy` in the repository root directory doesn't also check the example code.

.github/workflows/ci.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
         id: validation
         continue-on-error: true
         run: |
-          pushd examples
+          pushd example
```
kvcache, line 58: @cprice404 Does this address your `examples` folder concern for the way we are doing this action?

cprice404: renaming the dir to `example` is fine if we need to do that.  My only request is that the Cargo.toml file in the example dir specifies a version of momento that is pulled from crates.io so that our CI/dependabot can test the published packages for now, but i think you are doing that here.

kvcache (approved):
This looks good, thank you. We need to fix integration tests running from forks :-)

PR #91, opened 2023-01-24
feat: use impl IntoBytes instead of generics in certain cases

swlynch99:
This PR extends the use of `impl IntoBytes` from #88 to the rest of the functions within `SimpleCacheClient`.


swlynch99:
Seems like a typo in the title broke CI :(

kvcache (approved):
This modern rust impl trait parameter syntax reads more nicely, thanks.

Also, it has the side-effect of letting someone use a non-matched type for key and body in `set()` where previously they were required to be equivariant. That's a really nice addition 😄

malandis (approved):
Very nice!

PR #93, opened 2023-01-25
feat: remove redundant nested momento module

swlynch99:
Before this PR the auth module was in a nested momento module. So, to access it, you would need to look in `momento::momento::auth`. This is unnecessary and rather hard to use. This PR removes the nested momento module and drops the auth module into the crate root.


kvcache (approved):
thanks! This will mean a little fluff in `momento-cli` but it's better this way.

PR #94, opened 2023-01-25
feat: Convert the auth module to more idiomatic rust

swlynch99:
This PR is my attempt at making the `auth` module into a more idiomatic rust API.

The big changes are:
- Replace `LoginResult` with a `Result<Credentials, AuthError>`, error cases are moved to the error part of the result.
- `login` now has it's own custom error type to cover the custom error cases alongside regular errors. This type can be converted to `MomentoError` easily. This also means that we don't immediately convert all errors to strings, making things easier for users of the library to actually introspect what's going on.
- Rename `LoggedIn` -> `Credentials` and make the `valid_for_seconds` field a `Duration` instead. Also make fields private and instead use accessor methods.
- Delete `NotLoggedIn` since it is no longer used.

The internal changes are:
- Added `thiserror` as a dependency so that I can easily declare new error types.
- `utils::connect_channel_lazily` now returns an internal error type so that `?` works with it in functions returning both `MomentoError` and `AuthError`.
- I've also made it `pub(crate)` since it is not exposed publically.


kvcache (approved):
This is much nicer, thanks. We'll get @malandis or @pgautier404 to take a peek too, but this stuff is really only consumed in Rust at the moment so there's not a lot of consistency to maintain.

pgautier404 (approved):
LGTM!

malandis (dismissed):

swlynch99:
Rebased onto the latest main.

kvcache (approved):
applied pedantry
evident righteousnesses
thanks for the pr

PR #96, opened 2023-01-25
feat: Move `SimpleCacheClient` and associated types to crate root

swlynch99:
This PR makes the `simple_cache_client` module private and exports everything it used to export in the root module. `SimpleCacheClient` is the main type that users of the crate will be interacting with so it should really be in the root module. I have also re-exported `MomentoError` in the root module and added a `MomentoResult` type as a shorthand for `Result<T, MomentoError>`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,6 +7,9 @@ mod jwt;
 mod simple_cache_client;
 mod utils;
 
+pub use crate::response::MomentoError;
 pub use crate::simple_cache_client::{
     Fields, IntoBytes, SimpleCacheClient, SimpleCacheClientBuilder,
 };
+
+pub type MomentoResult<T> = Result<T, MomentoError>;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1,8 +1,15 @@
...
 
 mod endpoint_resolver;
 mod grpc;
 mod jwt;
+mod simple_cache_client;
 mod utils;
+
+pub use crate::response::error::MomentoError;
+pub use crate::simple_cache_client::{
+    request_meta_data, Fields, IntoBytes, SimpleCacheClient, SimpleCacheClientBuilder,
```
swlynch99, line 12: I'm not sure if `request_meta_data` is meant to be public since it looks like a helper function.

kvcache: nah, metadata should be an internal helper.

swlynch99: Cool. I've removed it from the export and made it private.

kvcache (approved):
I think this is much better, thank you. Flatter imports in a vacuum are better, and there's not a lot to disambiguate in the existing namespace.

PR #99, opened 2023-01-26
chore: Update example crate to use momento v0.15.0

swlynch99:
As in title.


kvcache (approved):
Thanks for the documentation bump

PR #101, opened 2023-01-26
feat: Make TTL parameters use more idiomatic Rust

swlynch99:
This PR does two main things:
- Use Duration for all time-span values with the exception of ttl_minutes when creating a signing key.
- Introduce a TtlPolicy type and use that for all TTL parameters in cache client methods where a TTL was previously taken. This is similar to the CollectionTtl type in the JS SDK but I've used it a bit more widely.

This means that all time-spans now go through std's Duration type and there is now now way that users could, for example, pass seconds to a method that is expecting milliseconds. The one exception to this for create_signing_key. Duration has no from_minutes method and I felt that having the duration silently truncate down to the last minute would be too unintuitive if the only constructor is Duration::from_secs.

One thing to note is that now users of the SDK can set TTLs to zero where that wasn't possible before. I consider this pretty minor since it was previously possible to set a TTL of 1ms which has effectively the same behaviour.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -1107,19 +1107,11 @@ impl SimpleCacheClient {
         Ok(())
     }
 
-    fn validate_ttl(&self, policy: &TtlPolicy) -> MomentoResult<Duration> {
-        let ttl = policy.ttl().unwrap_or(self.item_default_ttl);
+    fn expand_ttl_ms(&self, ttl: Option<Duration>) -> MomentoResult<u64> {
+        let ttl = ttl.unwrap_or(self.item_default_ttl);
         utils::is_ttl_valid(ttl)?;
-        Ok(ttl)
-    }
 
-    fn expand_ttl_ms(&self, policy: TtlPolicy) -> MomentoResult<u64> {
-        let ttl_ms = self
-            .validate_ttl(&policy)?
-            .as_millis()
-            .try_into()
-            .unwrap_or(u64::MAX);
-        Ok(ttl_ms)
+        Ok(ttl.as_millis().try_into().unwrap_or(i64::MAX as u64))
     }
 }
 
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -9,11 +9,11 @@ use momento_protos::{
 };
 use serde_json::Value;
 use std::collections::HashMap;
-use std::convert::TryFrom;
-use std::num::NonZeroU64;
+use std::convert::{TryFrom, TryInto};
+use std::time::Duration;
 use tonic::{codegen::InterceptedService, transport::Channel, Request};
 
-use crate::endpoint_resolver::MomentoEndpointsResolver;
+use crate::{endpoint_resolver::MomentoEndpointsResolver, MomentoResult};
 use crate::{grpc::header_interceptor::HeaderInterceptor, utils::connect_channel_lazily};
 
 use crate::response::{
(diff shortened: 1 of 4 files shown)


src/utils.rs
```diff
@@ -1,14 +1,16 @@
...
+    let max_ttl = Duration::from_millis(u64::MAX);
+    if ttl > max_ttl {
         return Err(MomentoError::InvalidArgument(format!(
             "TTL provided, {}, needs to be less than the maximum TTL {}",
-            ttl, max_ttl
+            ttl.as_secs(),
+            max_ttl.as_secs()
         )));
     }
     Ok(())
```
swlynch99, line 16: I have kept the validation for TTLs the same here. Would you consider instead just truncating the TTLs to u64::MAX instead of returning an error? I don't believe that truncating the max TTL to 554 billion years is likely to cause a problem for anyone :)

src/simple_cache_client.rs
```diff
@@ -168,11 +209,12 @@ impl SimpleCacheClientBuilder {
         );
         let data_client = ScsClient::new(data_interceptor);
 
+        #[allow(deprecated)]
```
kvcache, line 212: What is deprecated in here?

src/simple_cache_client.rs
```diff
@@ -41,13 +41,60 @@ where
     }
 }
 
+/// Policy controlling how the item TTL is updated when it is updated.
+#[derive(Copy, Clone, Debug)]
+pub struct TtlPolicy {
```
kvcache, line 46: This is going to need to be `CollectionTtl` for consistency https://github.com/momentohq/client-sdk-dotnet-incubating/blob/main/src/Momento.Sdk.Incubating/Requests/CollectionTtl.cs#L26

You might consider reusing some of the documentation in there too ❤️

src/simple_cache_client.rs
```diff
@@ -384,28 +425,28 @@ impl SimpleCacheClient {
     /// # Arguments
     ///
     /// * `cache_name` - name of cache
-    /// * `cache_key`
-    /// * `cache_body`
-    /// * `ttl_seconds` - If None is passed, uses the client's default ttl
+    /// * `cache_key` - key of entry within the cache.
+    /// * `cache_body` - data stored within the cache entry.
+    /// * `policy` - TTL policy to use. Note that set always sets the item TTL so refresh is ignored.
```
kvcache, line 430: The rename to `CollectionTtl` will make it more obvious, but this parameter should just be `Duration` to match our other sdks: https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/ISimpleCacheClient.cs#L130

src/simple_cache_client.rs
```diff
@@ -1093,6 +1107,21 @@ impl SimpleCacheClient {
...
+        utils::is_ttl_valid(ttl)?;
+        Ok(ttl)
+    }
+
+    fn expand_ttl_ms(&self, policy: TtlPolicy) -> MomentoResult<u64> {
+        let ttl_ms = self
+            .validate_ttl(&policy)?
+            .as_millis()
+            .try_into()
+            .unwrap_or(u64::MAX);
```
kvcache, line 1122: can you use `i64::MAX as u64` instead just so there's 1 less potential issue with any proxy we may have? It doesn't materially change the effect.

tests/integration_test.rs
```diff
@@ -14,14 +13,16 @@ mod tests {
...
+            Duration::from_secs(5),
             "integration_test",
             None,
         )
     }
 
     fn get_momento_instance() -> SimpleCacheClient {
-        let auth_token = env::var("TEST_AUTH_TOKEN").expect("env var TEST_AUTH_TOKEN must be set");
+        // let auth_token = env::var("TEST_AUTH_TOKEN").expect("env var TEST_AUTH_TOKEN must be set");
+        let auth_token = "eyJhbGciOiJIUzUxMiJ9.eyJzdWIiOiJzZWFuQG1vbWVudG9ocS5jb20iLCJjcCI6ImNvbnRyb2wuY2VsbC11cy1lYXN0LTEtMS5wcm9kLmEubW9tZW50b2hxLmNvbSIsImMiOiJjYWNoZS5jZWxsLXVzLWVhc3QtMS0xLnByb2QuYS5tb21lbnRvaHEuY29tIn0.hN-u3SCzFjV6wiStNkcHIATPQZNxHtW33_gKP6eo5Dlw4oYmCGTwJrsPZR9OW2mUvvj9G4uyUaR752AP9fSpPA"
```
kvcache, line 24: Woops.

kvcache (commented):
This is nice! There are a couple naming consistency things across our other sdks that we'd love to see addressed but this is definitely the right target shape for ttls in the Rust sdk.

tests/integration_test.rs
swlynch99: ah crap

src/simple_cache_client.rs
swlynch99: This was left over from a mid-refactoring change.

src/simple_cache_client.rs
```diff
@@ -41,13 +41,79 @@ where
...
+
+    /// Whether the collection's TTL will be refreshed on every update.
+    /// 
+    /// If true, this will extend the time at which the collection would expire when
+    /// an update operation happens. Otherwise, the collection's TTL will only be set
+    /// when it is initially created.
+    pub fn refresh(&self) -> bool {
+        self.refresh
+    }
+}
```
malandis, line 102: Looks good. While you're here, would you port the other methods from our standard? See [here](https://github.com/momentohq/client-sdk-dotnet-incubating/blob/main/src/Momento.Sdk.Incubating/Requests/CollectionTtl.cs) for reference

kvcache (approved):
looks great, thanks

PR #102, opened 2023-01-27
feat: add feature preview for topics

kvcache:
Momento topics are currently under development but this first step
helps enable end-user experimentation with the types.

This adds a Pubsub client, complete with connect(), publish() and
subscribe().

subscribe() is the more interesting method, as it returns a
Subscription, which is a new kind of user-facing functionality for
our sdk. I've wrapped the protobuf types into simpler non-Optional
messages, marshaling some of their safe/default notions into what
they mean for a Momento subscription.

For context, the subscription usage looks like:
```rust
let client = Pubsub::connect(token, endpoint, Some("github-demo"))?;

let subscription = client
    .subscribe(cache_name, topic, None)
    .await?;

while let Some(item) = subscription.item().await? {
    match item {
        SubscriptionItem::Value(value) => match value.kind {
            ValueKind::Text(text) => println!("{text}"),
            ValueKind::Binary(binary) => {
                println!("{{\"kind\": \"binary\", \"length\": {}}}", binary.len())
            }
        },
        SubscriptionItem::Discontinuity(discontinuity) => {
            println!("{discontinuity:?}")
        }
    }
}
```

add install_protoc.sh for tonic 0.8 in all the github files

also add dependency build cache for PR's. These approvals do not need to rebuild everything every time.

### DIFF
--- a/src/preview/topics.rs
+++ b/src/preview/topics.rs
@@ -1,0 +1,292 @@
+use momento_protos::cache_client::pubsub::{
+    self, pubsub_client::PubsubClient, PublishRequest, SubscriptionRequest,
+};
+use tonic::{codegen::InterceptedService, transport::Channel};
+
+use crate::MomentoError;
+use crate::{
+    endpoint_resolver::MomentoEndpointsResolver,
+    grpc::header_interceptor::HeaderInterceptor,
+    utils::{connect_channel_lazily, user_agent},
+    MomentoResult,
+};
+
+type ChannelType = InterceptedService<Channel, HeaderInterceptor>;
+
+pub struct Pubsub {
+    client: pubsub::pubsub_client::PubsubClient<ChannelType>,
+}
+
+/// Work with topics, publishing and subscribing.
+impl Pubsub {
+    pub fn connect(
+        auth_token: String,
+        momento_endpoint: Option<String>,
+        user_application_name: Option<&str>,
+    ) -> MomentoResult<Self> {
+        let momento_endpoints = MomentoEndpointsResolver::resolve(&auth_token, momento_endpoint)?;
+        let channel = connect_channel_lazily(&momento_endpoints.data_endpoint.url)?;
+        let authorized_channel = InterceptedService::new(
+            channel,
+            HeaderInterceptor::new(
+                &auth_token,
+                &user_agent(user_application_name.unwrap_or("sdk")),
+            ),
+        );
+        Ok(Self {
+            client: pubsub::pubsub_client::PubsubClient::new(authorized_channel),
+        })
+    }
+
+    /// Publish a value to a topic.
+    /// The cache is used as a namespace for your topics, and it needs to exist.
+    /// You don't create topics, you just start using them.
+    pub async fn publish(
+        &self,
+        cache_name: String,
+        topic: String,
+        value: impl IntoTopicValue,
+    ) -> Result<(), MomentoError> {
+        Pubsub::actually_publish(&mut self.client.clone(), cache_name, topic, value).await
+    }
+
+    /// Publish a value to a topic.
+    /// The cache is used as a namespace for your topics, and it needs to exist.
+    /// You don't create topics, you just start using them.
+    ///
+    /// Use this if you have &mut, as it will save you a small amount of overhead for reusing the client.
+    pub async fn publish_mut(
+        &mut self,
+        cache_name: String,
+        topic: String,
+        value: impl IntoTopicValue,
+    ) -> Result<(), MomentoError> {
+        Pubsub::actually_publish(&mut self.client, cache_name, topic, value).await
+    }
+
+    async fn actually_publish(
+        client: &mut PubsubClient<ChannelType>,
+        cache_name: String,
+        topic: String,
+        value: impl IntoTopicValue,
+    ) -> Result<(), MomentoError> {
+        client
+            .publish(PublishRequest {
+                cache_name,
+                topic,
+                value: Some(pubsub::TopicValue {
+                    kind: Some(value.into_topic_value()),
+                }),
+            })
+            .await?;
+        Ok(())
+    }
+
+    /// Subscribe to a topic.
+    /// The cache is used as a namespace for your topics, and it needs to exist.
+    /// You don't create topics, you just start using them.
+    pub async fn subscribe(
+        &self,
+        cache_name: String,
+        topic: String,
+        resume_at_topic_sequence_number: Option<u64>,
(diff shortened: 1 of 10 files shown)


src/preview/topics.rs
```diff
@@ -0,0 +1,292 @@
...
+};
+
+type ChannelType = InterceptedService<Channel, HeaderInterceptor>;
+
+pub struct Pubsub {
+    client: pubsub::pubsub_client::PubsubClient<ChannelType>,
+}
+
+/// Work with topics, publishing and subscribing.
+impl Pubsub {
```
cprice404, line 21: Just a naming question / quirk here.  Are we calling this feature Pubsub?  Or Topics?  Just want to try to make our branding consistent.

Also I think if we decide to go with PubSub as the outer name then the S should be capitalized.

src/preview/topics.rs
```diff
@@ -0,0 +1,292 @@
...
+    pub fn connect(
```
cprice404, line 22: Do you like having this as an explicit API vs handling it lazily in the publish or subscribe fns?

src/preview/topics.rs
```diff
@@ -0,0 +1,292 @@
...
+        auth_token: String,
+        momento_endpoint: Option<String>,
+        user_application_name: Option<&str>,
+    ) -> MomentoResult<Self> {
```
cprice404, line 26: If you do like having connect as an explicit API, and it's just returning Self, then would it make more sense to have it be a static factory method instead of requiring a constructor call followed by a connect call?

cprice404: oh, looking at your example code it kind of looks like that's already the pattern you are intending and I am just still very bad at reading rust code :(

src/preview/topics.rs
```diff
@@ -0,0 +1,292 @@
...
+        Ok(Self {
+            client: pubsub::pubsub_client::PubsubClient::new(authorized_channel),
+        })
+    }
+
+    /// Publish a value to a topic.
+    /// The cache is used as a namespace for your topics, and it needs to exist.
+    /// You don't create topics, you just start using them.
+    pub async fn publish(
+        &self,
```
cprice404, line 45: oh, right, so this is the thing that makes it instance method as opposed to static method?

src/preview/topics.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /// Publish a value to a topic.
+    /// The cache is used as a namespace for your topics, and it needs to exist.
+    /// You don't create topics, you just start using them.
+    ///
+    /// Use this if you have &mut, as it will save you a small amount of overhead for reusing the client.
+    pub async fn publish_mut(
+        &mut self,
+        cache_name: String,
```
cprice404, line 60: Ellery and I had been toying with the idea of calling this "namespace" instead of "cache" in the signatures for now until we can have a more thorough UX discussion.  Underlying implementation would be the same.  Just thinking about how thoroughly we want to permeate this choice throughout the APIs and codebases before we have actually had time to allow people to weigh in on UX.

src/preview/topics.rs
```diff
@@ -0,0 +1,292 @@
...
+        topic: String,
+        value: impl IntoTopicValue,
+    ) -> Result<(), MomentoError> {
```
cprice404, line 63: is it worth any treatment of what this error will look like if the connection is broken etc

cprice404 (approved):
handful of random questions / thoughts but no blockers.

src/preview/topics.rs
kvcache: I don't know that we actually have a decision on this. If we want to capitalize the S I'll just go TopicClient instead :-). I don't like the way PubSub reads. I'll update with TopicClient thanks.

src/preview/topics.rs
kvcache: this is the only function by which you can get a TopicClient (using my rename). I would not instantiate the Channel lazily though (like as a field of `Mutex<Box<ChannelType>>`), for a few reasons. I do like configuring the channel up front!

src/preview/topics.rs
kvcache: It's cool, I see lots of clients using static factory methods named like "connect" instead of a sterile "new" and I kind of like that. But the struct name I chose is not helpful in revealing that intent so yeah TopicClient::connect() for the win

src/preview/topics.rs
kvcache: yeah `&self` says "this function receives the self instance _as an immutable reference_"

src/preview/topics.rs
kvcache: I'm hesitant to name the variable other than cache_name at the moment, as it definitely requires that named cache. Calling it namespace in isolation here, without also documenting somewhere appropriate "what is a namespace" and educating our customers about a whole new abstract noun's meaning and semantics I think only serves to confuse this preview implementation.

I think calling it `namespace` now would be a mistake, but I think documenting it in signatures like "this is used as a topic namespace" makes good sense. I already wrote some copy for the `momento` cli to that effect (pr blocked on this one), to help ease folks into the word.

src/preview/topics.rs
kvcache: What do you have in mind? I'm reusing the same thing all our other apis use for error propagation (MomentoError) and it'll look the same as those.

cprice404: idk.  i just think this is a novel error case that i haven't seen with my own eyes yet, so i didn't know if there is anything specific that we should document about it.

kvcache: They all return Result<something, MomentoError> https://github.com/momentohq/client-sdk-rust/blob/main/src/simple_cache_client.rs#L418

src/preview/topics.rs
```diff
@@ -0,0 +1,315 @@
...
+                                // Broken protocol
+                                None => Some(SubscriptionItem::Discontinuity(Discontinuity {
+                                    last_sequence_number: None,
+                                    new_sequence_number: sequence_number,
+                                })),
+                            }
+                        }
+                        None => None, // Broken protocol
+                    },
+                    pubsub::subscription_item::Kind::Discontinuity(_) => todo!(),
```
PDXKimani, line 222: Is there some complexity here I'm not seeing? You already have the discontinuity types, isn't it a fairly straightforward conversion for this section? Any reason not to include it now?

kvcache: I missed this arm, good catch. It's addressed now.

PDXKimani (approved):

cprice404 (approved):

PR #103, opened 2023-01-27
chore: clean up internals of SimpleCacheClient

swlynch99:
This commit attempts to make the internals of `SimpleCacheClient` a bit more idiomatic. The changes it makes are:
- Replace `Result<T, MomentoError>` with `MomentoResult<T>` everywhere it is used.
- Move boilerplate around preparing a `tonic::Request` into a helper method and use that method wherever applicable.
- Replace all uses of `drain(..)` with `into_iter()` and extract this out into a helper method.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -850,7 +850,7 @@ impl SimpleCacheClient {
     /// use uuid::Uuid;
     /// use std::num::NonZeroU64;
     /// use std::env;
-    /// use momento::simple_cache_client::SimpleCacheClientBuilder;
+    /// use momento::SimpleCacheClientBuilder;
     ///
     /// let auth_token = env::var("TEST_AUTH_TOKEN").expect("TEST_AUTH_TOKEN must be set");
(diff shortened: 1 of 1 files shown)


src/simple_cache_client.rs
```diff
@@ -536,29 +537,30 @@ impl SimpleCacheClient {
...
+    ) -> MomentoResult<MomentoDictionarySetResponse> {
         let temp_ttl = ttl_seconds.unwrap_or(self.item_default_ttl_seconds);
         let ttl_to_use = match utils::is_ttl_valid(&temp_ttl) {
             Ok(_) => temp_ttl.get() * 1000_u64,
             Err(e) => return Err(e),
         };
-        let mut dictionary = dictionary;
         let items = dictionary
-            .drain()
+            .into_iter()
```
kvcache, line 547: It is more appropriate to use into_iter here but it is a bit subtle =)

kvcache (approved):
This all looks right, yes

PR #105, opened 2023-01-27
chore: Add support for SDK set difference command

swlynch99:
As in the title. I have also added aliases here to match those used in the .NET SDK - although I've still kept `set_union` and `set_difference` methods as the main ones with example documentation.

I've also made an attempt to make the doctest behave better in the face of assertion failures.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -26,8 +26,8 @@ use crate::response::{
     MomentoDictionaryFetchStatus, MomentoDictionaryGetResponse, MomentoDictionaryGetStatus,
     MomentoDictionaryIncrementResponse, MomentoDictionarySetResponse, MomentoDictionarySetStatus,
     MomentoError, MomentoGetResponse, MomentoGetStatus, MomentoListCacheResult,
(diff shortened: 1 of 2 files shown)


src/simple_cache_client.rs
```diff
@@ -1107,6 +1110,116 @@ impl SimpleCacheClient {
...
+                })),
+            },
+        )?;
+
+        let response = self.data_client.set_difference(request).await?.into_inner();
+
+        Ok(match response.set {
+            Some(Set::Found(_)) => FetchResponse::Found(()),
+            Some(Set::Missing(_)) | None => FetchResponse::Missing,
+        })
```
kvcache, line 1195: This isn't a "fetch" response. Maybe you want to name it `SetResponse` (gross) or something? Maybe @malandis or @pgautier404 have a thought here

swlynch99: My thought was that I could use this type across a bunch of different methods. I agree, though, that the name doesn't really match in this case. I'm not sure what a better one would be though.

swlynch99:
Marking this as draft for a bit until I can figure out some of the naming issues

PR #107, opened 2023-01-30
chore: clean up doc tests and avoid cache leaks on failure

swlynch99:
This PR cleans up the documentation so that
1. It reads nicer from within rustdoc
2. It properly cleans up the cache even in the cases where the test case panics or errors out.
3. It uses `?` instead of `.expect` everywhere. I've added `anyhow` so that if debugging a test case becomes a necessary it is still possible to get a backtrace to the line on which the error was generated.

This PR doesn't hit all the doc comments. I will add a follow-up one which finishes up the rest of the methods on `SimpleCacheClient`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,3 +28,5 @@ env_logger = "0.9.0"
 tokio = { version = "1.9.0", features = ["full"] }
 tokio-test = "0.4.2"
 uuid = { version = "0.8", features = ["v4"] }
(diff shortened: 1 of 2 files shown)


kvcache (approved):
This is nice, thanks. The pattern is repeated, so it would be nice & more readable to have it wrapped up in a helper (probably higher-order) function but this is still a positive step.

PR #109, opened 2023-01-31
feat: make set/delete responses return a custom empty type

swlynch99:
This PR attempts to make the set and delete methods on the client both more idiomatic as well as forward compatible with future changes. The actual changes are:
- The set methods used to always return a Status::OK enum, now they return an empty struct.
- The delete methods used to return `()`, now they also return an empty struct.

This way, if new data needs to be returned from these methods in the future it can be added directly to the return type without having to worry about making a breaking change.


src/response/mod.rs
```diff
@@ -13,11 +11,28 @@ mod list_signing_keys_response;
...
+                Self(())
+            }
+        }
+    };
+}
+
+empty_response!(MomentoSetResponse);
+empty_response!(MomentoDictionarySetResponse);
+empty_response!(MomentoDictionaryDeleteResponse);
+empty_response!(MomentoDeleteResponse);
```
kvcache, line 38: Instead of a new macro and a manually-defined default function, isn't is simpler to do

```rust
#[derive(Default)]
pub struct MomentoSetResponse {}

#[derive(Default)]
pub struct MomentoDeleteResponse {}

#[derive(Default)]
pub struct MomentoDictionarySetResponse {}

#[derive(Default)]
pub struct MomentoDictionaryDeleteResponse {}
```
?

Then below you do either `MomentoSetResponse::default()` or just `Default::default()`

swlynch99: - I don't want to have them derive `Default` since there's no guarantee that new fields added in the future will support default. The `new` function is only visible within the crate so we can change it in the future.
- If we define them like that then users can do `MomentoSetResponse {}` and we won't be able to add new fields at all. (`#[non_exhaustive]` could solve this though)

kvcache:
```rust
impl MomentoSetResponse { pub(crate) fn new() -> Self { Self {} } }
impl MomentoDeleteResponse { pub(crate) fn new() -> Self { Self {} } }
impl MomentoDictionarySetResponse { pub(crate) fn new() -> Self { Self {} } }
impl MomentoDictionaryDeleteResponse { pub(crate) fn new() -> Self { Self {} } }
```

is more in line with our simple caveperson style than a new macro =)

swlynch99: I have gone and fixed everything up to use the simpler style :)

kvcache (approved):

PR #111, opened 2023-01-31
chore: add a test-utils crate to make writing doctests easier

swlynch99:
As per some offline discussions, this refactors some of the common boilerplate in the doctests out to a utility function in a utility crate. It ended up being simpler than I thought it would be. I have also changed the doctest for `delete` to use this as an example.


kvcache (approved):
This is an awesome outcome of that earlier pr. Thanks for seeing it through! This way the doctests are cleaner and more relevant to users, and easier to correctly duplicate for other apis.

PR #112, opened 2023-01-31
feat: remove Fields enum and have dictionary_delete just take a vec

swlynch99:
`dictionary_delete` with `Fields::All` is exactly the same as just calling `delete`. Historically, there were reasons that there were different methods but those no longer apply. This PR removes the fields enum and makes `dictionary_delete` unconditionally take a `Vec<impl IntoBytes>`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,7 +11,7 @@ mod utils;
 pub use crate::response::ErrorSource;
 pub use crate::response::MomentoError;
 pub use crate::simple_cache_client::{
-    CollectionTtl, Fields, IntoBytes, SimpleCacheClient, SimpleCacheClientBuilder,
+    CollectionTtl, IntoBytes, SimpleCacheClient, SimpleCacheClientBuilder,
 };
 
 pub type MomentoResult<T> = Result<T, MomentoError>;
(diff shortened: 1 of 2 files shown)


malandis:
For context, in the wire format there is only one RPC `DictionaryDelete`, but in the SDK API elsewhere we surface three separate functions:
- `dictionary delete`: remove the entire dictionary from the cache
- `dictionary remove field`: remove a specific field from a particular dictionary
- `dicitionary remove fields`: remove multiple fields from a particular dictionary

The latter two were split out just as a convenience in some languages so users don't have to wrap/unwrap vectors of field names. So for forward compatibility I recommend renaming `dictionary_delete` to `dictionary_remove_fields` and consider adding `dictionary_delete`

swlynch99:
My contract with momento is up so @kvcache will probably have to do the follow up work here.

However, I can explain my thought process. Since momento deletes the dictionary entry once all its keys are removed calling `DictionaryDelete` with `Fields::All` is functionally the same as just deleting the key. My understanding was that previously the backend differentiated between the two so there was a reason to have a separate call to delete a dictionary but also that this is no longer the case.

malandis:
That's right! I forgot about that; thanks for the reminder. Disregard the above.

malandis:
This has been addressed in the newer cache client.

PR #113, opened 2023-01-31
chore: use test-utils crate in doctests and make examples more useful

swlynch99:
This PR extends the use of the `doctest` helper function from #111 to all doctests that interact with a specific cache. While going through the list I have also tried to make the doc tests more into examples that showcase behaviour rather than "here's one call to the function."


kvcache (approved):
This is very nice, thanks

PR #115, opened 2023-02-01
chore: implement all SDK list operations

swlynch99:
As in the title. I have included documentation and an example for each one. The one operation I implemented which was not in the protocol was `list_set` as a wrapper around `list_concat_front` since it made all the doctests much easier to understand.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -1382,6 +1382,28 @@ impl SimpleCacheClient {
     ///
     /// * `cache_name` - name of the cache in which the list is stored.
     /// * `list_name` - name of the list to pop from.
+    ///
+    /// # Example
+    /// ```
+    /// # fn main() -> momento_test_util::DoctestResult {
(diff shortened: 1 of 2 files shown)


src/simple_cache_client.rs
```diff
@@ -949,6 +950,617 @@ impl SimpleCacheClient {
...
+    pub async fn list_set<V: IntoBytes>(
+        &mut self,
+        cache_name: &str,
+        list_name: impl IntoBytes,
+        values: impl IntoIterator<Item = V>,
+        policy: CollectionTtl,
+    ) -> MomentoResult<()> {
+        // We're exposing this as a set so updating an existing entry should count as if
+        // we are creating a new entry.
+        let policy = policy.with_refresh_on_update();
```
kvcache, line 1256: Thoughtful touch.

kvcache (approved):
This looks good, thanks

PR #118, opened 2023-02-01
chore: Implement multiple-range list_erase variant

swlynch99:
This adds a `list_erase_many` which accepts multiple ranges. I didn't think that I would need this but then I immediately found a use case when looking at the redis commands so in it goes.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -1525,7 +1525,7 @@ impl SimpleCacheClient {
         list_name: impl IntoBytes,
         range: impl RangeBounds<u32>,
     ) -> MomentoResult<()> {
-        self.list_erase_many(cache_name, list_name, vec![range])
+        self.list_erase_many(cache_name, list_name, [range])
             .await
     }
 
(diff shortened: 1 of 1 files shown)


src/simple_cache_client.rs
```diff
@@ -1524,36 +1524,82 @@ impl SimpleCacheClient {
...
+    /// assert_eq!(values, expected);
+    /// # Ok(())
+    /// # })
+    /// # }
+    ///
+    pub async fn list_erase_many<R: RangeBounds<u32>>(
+        &mut self,
+        cache_name: &str,
+        list_name: impl IntoBytes,
+        ranges: Vec<R>,
```
kvcache, line 1569: do you need both vec and a bounded R here or could you use `&[impl Into<RangeBounds<u32>>]`?

swlynch99: You do need the bounded `R` since `RangeBounds` is a trait but there's no reason for it to have to be a `Vec`. I'll change it to an `impl IntoIterator<Item = R>`.

kvcache (approved):

PR #119, opened 2023-02-02
chore: revise error types

kvcache:
Adds special handling for the very very special no-error kinds of
http2 errors. This change digs into the root cause of some special
tonic statuses and exposes simple momento error kinds.

Also this change attaches `source`s to MomentoErrors, so users can
actually log Debug info that may be helpful. This is different from
the Display implementation, which is still high level / not very
helpful :-).


tests/integration_test.rs
```diff
@@ -50,18 +50,16 @@ mod tests {
         let mut mm = get_momento_instance();
         let result = mm.create_cache(cache_name).await.unwrap_err();
         let _err_msg = "Cache name cannot be empty".to_string();
-        assert!(matches!(
-            result,
-            MomentoError::InvalidArgument(_err_message)
-        ))
+        assert!(matches!(result.to_string(), _err_message))
```
swlynch99, line 53:
```suggestion
        assert_eq!(result.to_string(), "Cache name cannot be empty");
```
If the types are comparable then `assert_eq!` is better than doing `assert!(matches!(...))` because it will print out the values when it fails.

It looks like there's the same thing going on in a number of different test cases.

src/response/error.rs
```diff
@@ -1,70 +1,142 @@
...
+
 /// Exception type for resulting from invalid interactions with Momento Services.
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum MomentoError {
     /// Momento Service encountered an unexpected exception while trying to fulfill the request.
-    InternalServerError(String),
+    #[error("internal server error: {description}")]
+    InternalServerError {
+        description: String,
```
swlynch99, line 13: As a suggestion, use `Cow<'static, str>` here and in the other enum variants so that you can avoid allocating a string in most of the error cases.

src/response/error.rs
```diff
@@ -1,70 +1,142 @@
...
+    #[error("unknown source")]
+    Unknown(#[from] Box<dyn std::error::Error + Send + Sync>),
 
-impl From<http::uri::InvalidUri> for MomentoError {
-    fn from(e: http::uri::InvalidUri) -> Self {
-        // the uri gets derived from the jwt
-        Self::ClientSdkError(e.to_string())
-    }
-}
+    /// Caused by something in our backing library Tonic
```
swlynch99, line 117: This comment seems inaccurate. `AuthError` is from the `auth` module.

src/response/error.rs
```diff
@@ -74,26 +146,110 @@ impl From<tonic::Status> for MomentoError {
     }
 }
 
+// Status { code: Unknown, message: "error reading a body from connection: stream error received: not a result of an error", source: Some(hyper::Error(Body, Error { kind: Reset(StreamId(1), NO_ERROR, Remote) })) }
```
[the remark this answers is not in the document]
kvcache: not really. I'll drop it.

tests/integration_test.rs
kvcache: yeah, I agree - not sure why that was done that way. I don't want to conflate fixing that with this change though

src/response/error.rs
kvcache: mispaste, thanks

src/response/error.rs
kvcache: sure, nbd

swlynch99 (approved):
LGTM on my end. I really like this new error type :)

src/response/error.rs
```diff
@@ -1,70 +1,142 @@
...
+    TonicTransport(#[from] tonic::transport::Error),
+
+    /// Caused by something in our backing library Tonic
+    #[error("tonic status error")]
+    TonicStatus(#[from] tonic::Status),
+
+    /// Caused by a malformed URI
+    #[error("uri is invalid")]
+    InvalidUri(#[from] http::uri::InvalidUri),
 }
```
swlynch99, line 132: These should be using `#[source]` instead of `#[from]` right? We probably want to continue the source chain through the `ErrorSource`.

(this was meant to be part of the review above but I forgot to actually create the comment)

kvcache:
```
The Error trait’s source() method is implemented to return whichever field has a #[source] attribute or is named source, if any. This is for identifying the underlying lower level error that caused your error.

The #[from] attribute always implies that the same field is #[source], so you don’t ever need to specify both attributes.
```
`from` means `source + from` [according to the documentation](https://docs.rs/thiserror/latest/thiserror/)

PDXKimani (approved):

tests/integration_test.rs
```diff
@@ -50,18 +50,16 @@ mod tests {
...
 
     #[tokio::test]
     async fn key_id_validation() {
         let key_id = "";
         let mut mm = get_momento_instance();
         let result = mm.revoke_signing_key(key_id).await.unwrap_err();
-        assert!(matches!(result, MomentoError::InvalidArgument(_)))
+        let _expected = "".to_string();
+        assert!(matches!(result.to_string(), _expected))
     }
```
virratanasangpunth, line 63: > Adds special handling for the very very special no-error kinds of
http2 errors

What is this error case? Can we write an integ test for this to show how we're now handling it?

cprice404 (approved):
@kvcache this all looks fine to me relative to the time I had available to go over it.

I want to make sure you are aware of this:

https://github.com/momentohq/standards-and-practices/blob/main/docs/client-specifications/error-handling.md#error-types

That is a living doc and it doesn't have to be adhered to religiously, but I would like for our error types/messages/codes to be mostly consistent across languages.

I know you're handling some new/different types of errors here and I trust your judgment but just wanted to make sure you knew about that doc so you could decide which parts of it are applicable, and maybe help me update the doc if there are parts that need to be tweaked.

tests/integration_test.rs
kvcache: I don't know how to instigate it in an integration test Vir. Server reset with no error and h2 goaway are the known causes currently. Read the documentation and/or instantiation site for the MomentoError::Interrupted enumeration for more details. If you'd like to dive deeper you can read about NO_ERROR in `h2` https://github.com/hyperium/h2/blob/07d20b19abfd3bf57bd80976089e3c24a3166bca/src/proto/connection.rs#L217-L219 and the cases in which it is the reason for `h2` family errors.

kvcache:
@cprice404 thanks! I'm trying to color in the weak semantic errors we had (just strings) with their causes (like putting Exception(message, cause) in java rather than Exception(message) and dropping the cause context).

The 1 thing I'm intending to _add_ to the semantics set is `Interrupted`. It's a non-error case that is exceptional, but not indicative of a problem necessarily. It's important for Topic subscriptions, as eventually no matter what we're going to need to deploy something & the connection will be Interrupted. This is not an error, but it follows an interruption pattern that Error is suitable for.

I'm not trying to improve the compliance of Rust errors with S&P with this change, but the only decrease in compliance I intend is `Interrupted`. It might be worth talking about this more in detail on Slack or Zoom? I suspect Interrupted should have a home in S&P.

PR #124, opened 2023-02-14
feat: pin to working protos version, add lockfile

cprice404:
A patch release of the protos contained a breaking change, and
because we didn't pin to a specific version or have a Cargo.lock
file, this caused the build to break for this project and
consuming projects.  This commit adds the Cargo.lock file,
and (for now) pins to a version of the client protos that this
library will compile successfully against.


Cargo.toml
```diff
@@ -18,7 +18,7 @@ members = [
 exclude = [ "example" ]
 
 [dependencies]
-momento-protos = { version = "0.42" }
+momento-protos = { version = "=0.42.4" }
```
kvcache, line 21: I don't really think it makes sense to pin to explicit versions like this. Semantically the SDK depends on momento protos 0.42, and the lockfile is the better tool for explicit versions.

I'm not going to block you on this and I will approve once the approvals are green but I don't agree with this style.

cprice404: right now if i don't do this, it literally won't compile because 0.42.5 is not compatible.

kvcache: you could put the version lock in the new cargo.lock file like this

```
[[package]]
name = "momento-protos"
version = "0.42.4"
```

cprice404:
@kvcache i pushed a commit that attempts to remove the chrono dep.

Cargo.toml
kvcache: One can edit any file one wishes to edit. It's certainly atypical, but it's also quite atypical to publish a library without a cargo.lock file :-(

As I said, I'm not going to block you but I disagree with this plan.

src/simple_cache_client.rs
```diff
@@ -395,11 +394,7 @@ impl SimpleCacheClient {
         let response = MomentoCreateSigningKeyResponse {
             key_id: kid.as_str().expect("'kid' not a valid str").to_owned(),
             key: res.key,
-            expires_at: DateTime::<Utc>::from_utc(
-                NaiveDateTime::from_timestamp_opt(res.expires_at as i64, 0)
-                    .expect("couldn't parse from timestamp"),
-                Utc,
-            ),
+            expires_at: SystemTime::from(UNIX_EPOCH + Duration::from_secs(res.expires_at)),
```
kvcache, line 397: This is nice.

kvcache (commented):
Thanks for removing that chrono dependency! We'll figure out why the approval isn't running or I'll pull this down and independently verify tomorrow morning.

brayniac:
I'm under the impression the cargo lock file in a library is ignored. So
your users will still have an issue if you don't explicitly pin in the
manifest

kvcache:
> I'm under the impression the cargo lock file in a library is ignored. So your users will still have an issue if you don't explicitly pin in the manifest

That would be news to me, but I don't have a ton of cargo distribution experience. Thanks for your insight!

brayniac:
Specifically, I think it'll work in the library when you build the library.
But it is not honored by consumers of your library. It seems to still be
advised to not commit a lock file in a library only project. Was just
chatting with Sean about this the other day for another project

kvcache:
> Specifically, I think it'll work in the library when you build the library. But it is not honored by consumers of your library. It seems to still be advised to not commit a lock file in a library only project. Was just chatting with Sean about this the other day for another project

Okay, so re-reading documentation on this yet again it appears I was _exactly_ backwards on what I remembered and confidently gave the opposite guidance of the correct best practice. 🤦 Thank you.

I remembered this page clearly and opposite of what it is in reality: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries We should not check in a lockfile to this repo.

cprice404:
I've put the lock file back into .gitignore.

kvcache (approved):

PR #126, opened 2023-02-15
chore: Handle heartbeats in topics

kvcache:
This adds support for heartbeats that will be rolled out more broadly
soon. We'll need to be able to accept these messages and allow the
subscription to continue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ members = [
 exclude = [ "example" ]
 
 [dependencies]
-momento-protos = { version = "=0.42.4" }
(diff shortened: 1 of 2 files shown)


PDXKimani (approved):

tylerburdsall (commented):
Clippy is suggesting some fixes, but once those are addressed, let's get this merged

tylerburdsall (approved):
Docs test are failing but that isn't new iirc

kvcache:
> Docs test are failing but that isn't new iirc

yeah that's not new. There's something wrong with the test runner + the remote test dependency

PR #129, opened 2023-02-21
feat: add support for flush cache API

cprice404:
I have some thoughts on tweaking some of the APIs in this repo a bit before 1.0 now that I have gotten to look at them a bit, so this might not be the final API for flush cache; but this unblocks us for getting it into the CLI, which is something customers are wanting / needing, so I would like to move forward with this API for now.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -1,3 +1,4 @@
+use momento_protos::control_client::FlushCacheRequest;
 use momento_protos::{
     cache_client::scs_client::*,
     cache_client::*,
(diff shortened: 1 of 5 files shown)


gautamomento (approved):

src/simple_cache_client.rs
```diff
@@ -368,6 +370,18 @@ impl SimpleCacheClient {
         Ok(response)
     }
 
+    pub async fn flush_cache(
```
kvcache, line 373: It would be very nice to have a docstring on this describing its semantics and call rate; ideally a doctest too so folks on docs.rs can be sure they're reading up-to-date examples like many of the other apis.

kvcache (commented):
looks good

PR #132, opened 2023-03-03
feat: add new generate api token api

bruuuuuuuce:
Adds a new api to generate expirable tokens

Corresponding client-protos https://github.com/momentohq/client-protos/blob/main/proto/controlclient.proto#L89-L116

this commit also contains some fixes to both the build as well as our integration tests

- Builds not happen in a `macos-latest` environment on github. We were running into flaky broken pipe issues within the ubuntu environment
- integration/doc test caches are now prefixed with `rust-sdk`
- updated protoc to vesion `21.9`


src/simple_cache_client.rs
```diff
@@ -1917,35 +1918,7 @@ impl SimpleCacheClient {
...
-            .sorted_set_fetch(request)
-            .await?
-            .into_inner();
-        Ok(MomentoSortedSetFetchResponse {
-            value: response.sorted_set.and_then(|s| match s {
-                SortedSet::Found(found) => Some(found.elements),
-                SortedSet::Missing(_) => None,
-            }),
-        })
+        todo!();
```
bruuuuuuuce, line 1921: this api is getting overhauled. When I updated the client protos, it was breaking this api. Since this api is still getting hashed out, I decided to just leave it as a `todo()`

src/simple_cache_client.rs
```diff
@@ -1991,13 +1964,15 @@ impl SimpleCacheClient {
...
+        use momento_protos::cache_client::sorted_set_get_rank_request::Order::Ascending;
         use momento_protos::cache_client::sorted_set_get_rank_response::Rank;
 
         let request = self.prep_request(
             cache_name,
             SortedSetGetRankRequest {
                 set_name: set_name.into_bytes(),
-                element_name: element_name.into_bytes(),
+                value: element_name.into_bytes(),
+                order: Ascending.into(),
```
bruuuuuuuce, line 1975: This is not changing the behavior of this method at all. The `Ascending` enum has a value of 0, which is the default value that the proto would have. Since its now required by the `SortedSetGetRankRequest`, I am just passing in what is the default value. We need to come back into this sdk and rework the sorted set methods.

```
message _SortedSetGetRankRequest {
  enum Order {
    ASCENDING = 0;
    DESCENDING = 1;
  }
```

.github/workflows/cd.yml
```diff
@@ -42,15 +42,15 @@ jobs:
         run: echo "::set-output name=release::${{ steps.semrel.outputs.version }}"
 
   publish:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
cprice404, line 45: i'm not sure how long this is going to fly for; we may need to figure out how to get this running on ubuntu-20.04 or dig deeper into the networking issues that are causing the failures on 22.04, but for now I think we can just keep an eye on the bill.

src/simple_cache_client.rs
```diff
@@ -1873,79 +1874,20 @@ impl SimpleCacheClient {
...
-    ///         for entry in &set {
-    ///             println!("{} {:?}", entry.score, entry.name);
-    ///         }
-    ///     },
-    ///     None => println!("sorted set not found!"),
-    /// }
-    /// # Ok(())
-    /// # })
-    /// # }
-    /// ```
```
cprice404, line 1911: was this stuff removed intentionally?

src/simple_cache_client.rs
```diff
@@ -1873,79 +1874,20 @@ impl SimpleCacheClient {
...
-        cache_name: &str,
-        set_name: impl IntoBytes,
-        order: sorted_set::Order,
-        limit: impl Into<Option<NonZeroU32>>,
-        range: Option<sorted_set::Range>,
+        _cache_name: &str,
+        _set_name: impl IntoBytes,
+        _order: sorted_set::Order,
+        _limit: impl Into<Option<NonZeroU32>>,
+        _range: Option<sorted_set::Range>,
```
cprice404, line 1888: what's up with these name changes?  This is a public API, right?

src/simple_cache_client.rs
cprice404: oh, I see.  The API is settled now.  The support for the final API has already been added to the node.js SDK.

cprice404: who originally added SortedSet to this SDK?  we need to check and see if it is being used by someone.  If not, maybe this is okay for now.  cc: @kvcache

test-util/src/lib.rs
```diff
@@ -24,7 +24,7 @@ where
     // The constructor for the cache client needs a tokio runtime to be active.
     let _guard = runtime.enter();
 
-    let cache_name = Uuid::new_v4().to_string();
+    let cache_name = "rust-sdk-".to_string() + &Uuid::new_v4().to_string();
```
cprice404, line 27: THANK YOU

cprice404 (approved):
I'm mostly 🙈 on this but @kvcache should weigh in on the sorted set parts.

src/response/generate_api_token_response.rs
```diff
@@ -0,0 +1,17 @@
+use std::time::SystemTime;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ApiKeyWithEndpoint {
+    pub api_key: String,
+    pub endpoint: String,
+}
```
kvcache, line 9: You should not derive Debug here - you should impl it and redact the api_key, otherwise this will show up in logs incidentally.

src/response/generate_api_token_response.rs
```diff
@@ -0,0 +1,17 @@
...
+
+/// The result of a generate api token operation.
+#[derive(Debug)]
+pub struct MomentoGenerateApiTokenResult {
+    pub api_token: String,
```
kvcache, line 14: same here about deriving debug

src/response/generate_api_token_response.rs
```diff
@@ -0,0 +1,17 @@
+use std::time::SystemTime;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ApiKeyWithEndpoint {
+    pub api_key: String,
```
kvcache, line 7: What's the difference between an api key and an api token?

src/response/generate_api_token_response.rs
```diff
@@ -0,0 +1,17 @@
...
+    pub endpoint: String,
+}
+
+/// The result of a generate api token operation.
+#[derive(Debug)]
+pub struct MomentoGenerateApiTokenResult {
```
kvcache, line 13: In Rust, `Result` evokes thoughts of the `Result` type. Often times people will write their own Error type and define a *Result to wrap it up.

By my read in 9 out of 10 of the `_response` files in this `response` module, the response type is called a `Response`. Is there some reason I'm missing to keep this named "result?"

src/simple_cache_client.rs
kvcache: I do not know whether this is consumed, but @brayniac is the most likely person I am aware of to be using it.

Is there an issue we can link here instead like
```
todo!("This api was reworked and is pending implementation in the rust sdk: https://github.com/foo/bar");
```
instead of a contextless panic?

src/simple_cache_client.rs
```diff
@@ -2312,6 +2256,36 @@ impl SimpleCacheClient {
         Ok(MomentoDeleteResponse::new())
     }
 
+    /// Generates a token for Momento
```
kvcache, line 2259: Consistency in verbiage will help with the clarity of these functions: I see you seem to use "token," "api token" and "api key" interchangeably (I think?) in this pr. We've also got a "session token" for sso so this isn't getting any clearer!

src/simple_cache_client.rs
bruuuuuuuce: since this function is now empty, these variables are not used. Clippy wants unused variables to start with `_`

cprice404: yeah i get it now.  it's fine pending resolution of the sortedset convo below.

src/response/generate_api_token_response.rs
bruuuuuuuce: We had this discussion in the design meeting. Basically we are calling the base64 encoded `(endpoint, api_key)` the `api_token`. We thought that it made the most sense because that string they will just treat as a token, and there is no need for them to base64 decode that string. A user should only have to deal with `_tokens`. So `refresh_token`, `api_token`, `session_token`.

bruuuuuuuce: This also matches what is in the protos

src/response/generate_api_token_response.rs
bruuuuuuuce: good call

src/response/generate_api_token_response.rs
kvcache: Good point, probably undesirable to print any secret material as a side-effect of a well-intentioned log::info. I'd recommend not printing refresh token (or printing `refresh_token: "<redacted>"` to help keep it friendly and clear where it lives)

src/response/generate_api_token_response.rs
kvcache: Should this struct be called ApiToken with a `/// Comment` to that effect?

bruuuuuuuce: yea I can change the name and add a comment

src/simple_cache_client.rs
bruuuuuuuce: yep thats correct

bruuuuuuuce: resolved in https://github.com/momentohq/client-sdk-rust/pull/132/commits/a10baf2df4d874788a9eae80e49a54fbedf939bd

src/response/generate_api_token_response.rs
bruuuuuuuce: fixed in https://github.com/momentohq/client-sdk-rust/pull/132/commits/a10baf2df4d874788a9eae80e49a54fbedf939bd

src/response/generate_api_token_response.rs
bruuuuuuuce: added comment in this commit https://github.com/momentohq/client-sdk-rust/pull/132/commits/a10baf2df4d874788a9eae80e49a54fbedf939bd

src/response/generate_api_token_response.rs
bruuuuuuuce: fixed in https://github.com/momentohq/client-sdk-rust/pull/132/commits/a10baf2df4d874788a9eae80e49a54fbedf939bd.

src/response/generate_api_token_response.rs
bruuuuuuuce: fixed in https://github.com/momentohq/client-sdk-rust/pull/132/commits/a10baf2df4d874788a9eae80e49a54fbedf939bd

cprice404 (approved):

kvcache (approved):

PR #136, opened 2023-03-06
fix: publish Expires, Expiry, Never from generate token protos

bruuuuuuuce:
The `generate_api_token` request takes in an `Expiry` enum. This enum comes directly from the client protos, and without publishing the enum and its values, a customer cannot call the `generate_api_token`.

https://github.com/momentohq/client-sdk-rust/blob/main/src/simple_cache_client.rs#L2266

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,6 @@
 pub mod auth;
 pub mod preview;
+pub mod requests;
 pub mod response;
 
 mod endpoint_resolver;
@@ -19,10 +20,4 @@ pub type MomentoResult<T> = Result<T, MomentoError>;
 pub mod sorted_set {
     pub use momento_protos::cache_client::sorted_set_fetch_request::{Order, Range};
     pub use momento_protos::cache_client::SortedSetElement;
-}
-
-pub mod token_expiry {
-    pub use momento_protos::control_client::generate_api_token_request::Expires;
-    pub use momento_protos::control_client::generate_api_token_request::Expiry;
-    pub use momento_protos::control_client::generate_api_token_request::Never;
 }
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -20,3 +20,9 @@ pub mod sorted_set {
     pub use momento_protos::cache_client::sorted_set_fetch_request::{Order, Range};
     pub use momento_protos::cache_client::SortedSetElement;
 }
+
+pub mod tokens {
```
bruuuuuuuce, line 24: not sure if people have strong opinions on what this name should be

cprice404: maybe call it `token_expiry` and we can limit it to just this enum of values?  Just thinking we might wanna use token for something else in the future.  but this is also internal, 2-way door stuff that doesn't appear in any user-facing APIs, right?

bruuuuuuuce: > maybe call it token_expiry

I like that

> we can limit it to just this enum of values?

Yep

>  but this is also internal, 2-way door stuff that doesn't appear in any user-facing APIs, right?

the `generate_api_token` method signature takes in an `Expiry` enum value

bruuuuuuuce, line 24:
```suggestion
pub mod token_expiry {
```

cprice404: in that case I think you need to create a custom enum directly in the SDK and map it.  You don't want to allow generated types (or generally, any types that are defined by one of your dependencies) leak through into your public API unless there's a really good reason for it.

cprice404: cc: @kvcache

bruuuuuuuce: ah ok, I was trying to follow the pattern set by sorted sets, but that reasoning makes sense to me

cprice404: @kvcache lmk if you disagree on this but that's the general rule of thumb I have followed for this sort of thing over the years

cprice404 (approved):
i think this is better personally :)

src/lib.rs
kvcache: > unless there's a really good reason for it

I have _rarely_ found a good reason for it. They exist but usually only because of gaps in build tools as opposed to some inherent reason in a domain model...

kvcache (approved):

PR #142, opened 2023-04-14
feat: structure response::Get

kvcache:
There are a couple updates in here.
First up, I'm removing the momento::response::MomentoGetStatus
enum entirely.

Next, I'm turning momento::response::MomentoGetResponse into an
enum. In the migration I'm removing "ERROR" from the enumerated
response states. We have MomentoResult, which encapsulates Error
already. Doubling the places to check for errors is not good!

In addition to the removal, I'm changing the naming convention
from scream case to the more normal Rust enum title case.

While looking at the names, I observed that there is no difference
in the ambiguity of these 2 naming options:
* momento::response::MomentoGetResponse
* momento::response::Get

however, the latter is much lighter syntactically and offers
familiar options for importing at whatever level of disambiguation
one might desire. It also does not repeat itself.

Rather than migrating the as_string() function of MomentoGetResponse,
which was actually more of an as_str(), I've implemented a couple of
TryFrom specializations. These are for users to easily cast responses
as hits in String and Vec<u8> and let errors propagate.
This allows people to choose if they want to treat Miss as Error.

In deciding not to migrate as_string(), I'm also closing off an
implicit behavior we had, where a non-utf-8 string would have been
silently treated as a Hit( "" ). Rather than clobber cache items,
the new response::Get's TryInto<String> and &str implementations
return a new momento::MomentoError::TypeError with the associated
cause wired up in the Rust way for nice tracebacks.

For users that match on hit/miss, I have added a GetValue for which
String has a TryFrom implementation and Vec<u8> is covered via From.
Documentation on the momento::response::Get demonstrates options for
parsing our responses, depending on the user's needs. I've included
doctests to help users know what their easy options are.


src/response/cache_dictionary_fetch_response.rs
```diff
@@ -1,20 +1,56 @@
...
+/// Response for a dictionary fetch operation.
+#[derive(Debug, PartialEq, Eq)]
+pub enum DictionaryFetch {
+    /// The dictionary item existed in the cache
+    Hit { value: DictionaryPairs },
+    /// The dictionary item did not exist
+    Miss,
+}
+
+impl TryFrom<DictionaryFetch> for HashMap<Vec<u8>, Vec<u8>> {
```
cprice404, line 19: did you decide you want to keep this?  I'm still kinda in favor of removing this one and just supporting strings for map keys for now

src/response/cache_dictionary_get_response.rs
```diff
@@ -1,20 +1,56 @@
...
-    pub dictionary: Option<HashMap<Vec<u8>, Vec<u8>>>,
+#[derive(Debug, PartialEq, Eq)]
+pub enum DictionaryGet {
+    /// The dictionary item existed in the cache
+    Hit { value: DictionaryPairs },
+    /// The dictionary item did not exist
+    Miss,
+}
+
+impl TryFrom<DictionaryGet> for HashMap<Vec<u8>, Vec<u8>> {
```
cprice404, line 19: same comment here

src/response/cache_get_response.rs
```diff
@@ -1,25 +1,106 @@
...
+/// ```
+/// # use momento::response::Get;
+/// # use momento::MomentoResult;
+/// # let get_response = Get::Hit { value: momento::response::GetValue::new(vec![]) };
+/// let item: Vec<u8> = match get_response {
+///     Get::Hit { value } => value.into(),
+///     Get::Miss => return // probably you'll do something else here
+/// };
+/// ```
+/// Or, if you are using utf-8 strings:
```
cprice404, line 19: i would put the string one first

cprice404 (commented):
couple of nits.  they are probably not blockers.

src/response/cache_get_response.rs
kvcache: yep

src/response/cache_dictionary_fetch_response.rs
kvcache: will drop

src/response/cache_dictionary_get_response.rs
kvcache: will drop

cprice404 (approved):
:ty:

PR #144, opened 2023-05-13
feat: extend topic Subscription to support Stream

kvcache:
futures::Stream is a better integration point for topics than a
one-off async method.

This change also makes Subscriptions try really hard to reconnect.
As long as you subscribe and await another item, the Subscription
will drive toward reconnecting and continuing.


cprice404 (approved):
🙈

PR #145, opened 2023-06-05
fix: disable integration tests

kvcache:
Ongoing auth work has updated the token used by this test runner.
These tests no longer work and block updates to the sdk. Their
asserts should be restored in one way or another, but they need
to not block publish.

Also sneaking in a small update to the keepalive ping timer to help
clarify debug log messages on momento-cli.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -167,17 +167,18 @@ impl SimpleCacheClientBuilder {
     /// * `default_ttl` - Default TTL for items put into a cache.
     /// # Examples
     ///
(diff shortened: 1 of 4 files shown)


tylerburdsall (dismissed):

rlinehan:
Looks good but also looks like there's some more tests that need to be disabled?

kvcache:
@rlinehan I think I got them all now...

rlinehan (approved):
Thanks! I'll reenable these in the PR I'm working on to add v1 token support.

PR #148, opened 2023-07-03
feat: add sorted_set_fetch_by_index

brayniac:
Adds a function to the simple cache client to fetch elements of a sorted set selecting them by index (rank).

### DIFF
--- a/src/response/cache_sorted_set_fetch_response.rs
+++ b/src/response/cache_sorted_set_fetch_response.rs
@@ -1,7 +1,7 @@
-use momento_protos::cache_client::SortedSetElement;
+use momento_protos::cache_client::sorted_set_fetch_response::SortedSet;
 
 #[derive(Debug)]
 #[non_exhaustive]
 pub struct MomentoSortedSetFetchResponse {
-    pub value: Option<Vec<SortedSetElement>>,
+    pub value: Option<SortedSet>,
 }
--- a/src/response/cache_sorted_set_fetch_response.rs
+++ b/src/response/cache_sorted_set_fetch_response.rs
@@ -1,7 +1,147 @@
+use crate::sorted_set::Elements;
+use crate::MomentoError;
+use core::convert::TryFrom;
 use momento_protos::cache_client::sorted_set_fetch_response::SortedSet;
 
 #[derive(Debug)]
 #[non_exhaustive]
 pub struct MomentoSortedSetFetchResponse {
     pub value: Option<SortedSet>,
+}
+
+pub enum SortedSetFetch {
+    Found { elements: Elements },
+    Missing,
+}
+
+impl TryFrom<SortedSetFetch> for Vec<(Vec<u8>, f64)> {
+    type Error = MomentoError;
+
+    fn try_from(value: SortedSetFetch) -> Result<Self, Self::Error> {
+        match value {
+            SortedSetFetch::Found { elements } => match elements {
+                Elements::ValuesWithScores(mut values) => {
+                    let result: Vec<(Vec<u8>, f64)> = values
+                        .elements
+                        .drain(..)
+                        .map(|e| (e.value, e.score))
+                        .collect();
+                    Ok(result)
+                }
+                Elements::Values(_) => Err(MomentoError::TypeError {
+                    description: std::borrow::Cow::Borrowed(
+                        "response did not contain element scores",
+                    ),
+                    source: Box::new(std::io::Error::new(
+                        std::io::ErrorKind::InvalidData,
+                        "element did not contain a score",
+                    )),
+                }),
+            },
+            SortedSetFetch::Missing => Err(MomentoError::Miss {
+                description: std::borrow::Cow::Borrowed("sorted set was not found"),
+            }),
+        }
+    }
+}
+
+impl TryFrom<SortedSetFetch> for Vec<Vec<u8>> {
+    type Error = MomentoError;
+
+    fn try_from(value: SortedSetFetch) -> Result<Self, Self::Error> {
+        match value {
+            SortedSetFetch::Found { elements } => match elements {
+                Elements::ValuesWithScores(mut values) => {
+                    let result: Vec<Vec<u8>> = values.elements.drain(..).map(|e| e.value).collect();
+                    Ok(result)
+                }
+                Elements::Values(values) => Ok(values.values),
+            },
+            SortedSetFetch::Missing => Err(MomentoError::Miss {
+                description: std::borrow::Cow::Borrowed("sorted set was not found"),
+            }),
+        }
+    }
+}
+
+impl TryFrom<SortedSetFetch> for Vec<(String, f64)> {
+    type Error = MomentoError;
+
+    fn try_from(value: SortedSetFetch) -> Result<Self, Self::Error> {
+        match value {
+            SortedSetFetch::Found { elements } => match elements {
+                Elements::ValuesWithScores(mut values) => {
+                    let mut result = Vec::with_capacity(values.elements.len());
+                    for element in values.elements.drain(..) {
+                        let value = String::from_utf8(element.value).map_err(|e| {
+                            return Err::<Self, Self::Error>(MomentoError::TypeError {
+                                description: std::borrow::Cow::Borrowed(
+                                    "element value was not a valid utf-8 string",
+                                ),
+                                source: Box::new(e),
+                            });
+                        });
+                        result.push((value.unwrap(), element.score));
+                    }
+                    Ok(result)
+                }
+                Elements::Values(_) => Err(MomentoError::TypeError {
+                    description: std::borrow::Cow::Borrowed(
+                        "response did not contain element scores",
+                    ),
+                    source: Box::new(std::io::Error::new(
+                        std::io::ErrorKind::InvalidData,
+                        "element did not contain a score",
+                    )),
+                }),
+            },
+            SortedSetFetch::Missing => Err(MomentoError::Miss {
+                description: std::borrow::Cow::Borrowed("sorted set was not found"),
+            }),
+        }
+    }
+}
+
+impl TryFrom<SortedSetFetch> for Vec<String> {
+    type Error = MomentoError;
+
+    fn try_from(value: SortedSetFetch) -> Result<Self, Self::Error> {
+        match value {
+            SortedSetFetch::Found { elements } => match elements {
+                Elements::ValuesWithScores(mut values) => {
+                    let mut result = Vec::with_capacity(values.elements.len());
+                    for element in values.elements.drain(..) {
+                        let value = String::from_utf8(element.value).map_err(|e| {
+                            return Err::<Self, Self::Error>(MomentoError::TypeError {
+                                description: std::borrow::Cow::Borrowed(
+                                    "element value was not a valid utf-8 string",
+                                ),
+                                source: Box::new(e),
+                            });
+                        });
+                        result.push(value.unwrap());
+                    }
+                    Ok(result)
+                }
+                Elements::Values(mut values) => {
+                    let mut result = Vec::with_capacity(values.values.len());
+                    for value in values.values.drain(..) {
+                        let value = String::from_utf8(value).map_err(|e| {
+                            return Err::<Self, Self::Error>(MomentoError::TypeError {
+                                description: std::borrow::Cow::Borrowed(
+                                    "element value was not a valid utf-8 string",
+                                ),
+                                source: Box::new(e),
+                            });
+                        });
+                        result.push(value.unwrap());
+                    }
+                    Ok(result)
+                }
+            },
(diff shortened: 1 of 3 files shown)


src/response/cache_sorted_set_fetch_response.rs
```diff
@@ -1,7 +1,7 @@
-use momento_protos::cache_client::SortedSetElement;
+use momento_protos::cache_client::sorted_set_fetch_response::SortedSet;
 
 #[derive(Debug)]
 #[non_exhaustive]
 pub struct MomentoSortedSetFetchResponse {
```
brayniac, line 5: I'm unsure if we really want this wrapper type. It doesn't appear to provide much value?

src/simple_cache_client.rs
```diff
@@ -1873,6 +1873,133 @@ impl SimpleCacheClient {
...
+
+        let request = self.prep_request(
+            cache_name,
+            SortedSetFetchRequest {
+                set_name: set_name.into_bytes(),
+                order: order.into(),
+                with_scores,
+                range: Some(sorted_set_fetch_request::Range::ByIndex(
+                    sorted_set_fetch_request::ByIndex {
+                        start: Some(start),
```
brayniac, line 1978: Is there any downside to always providing Some start and Some end? I'm confused why we have an Option and then either some bound or the dedicated Unbounded type.

src/simple_cache_client.rs
```diff
@@ -1873,6 +1873,137 @@ impl SimpleCacheClient {
...
+            },
+        )?;
+
+        let response = self
+            .data_client
+            .sorted_set_fetch(request)
+            .await?
+            .into_inner();
+
+        // this flattens the response returning a None for both a missing sorted
```
brayniac, line 1991: I'm not sure if this is the behavior we want, but the ergonomics feel funny otherwise. Do we ever really return a Found<None> instead of a Found<Some(elements)> where elements contains a zero length vector of elements? If so, we do lose information here, but otherwise I think this flattening is beneficial.

src/simple_cache_client.rs
```diff
@@ -1873,6 +1873,137 @@ impl SimpleCacheClient {
...
+    /// match momento.sorted_set_fetch_by_index(
+    ///     &cache_name,
+    ///     "test sorted set",
+    ///     Order::Ascending,
+    ///     0..10,
+    ///     true,
+    /// ).await? {
+    ///     Some(elements) => match elements {
+    ///         Elements::ValuesWithScores(elements) => {
+    ///             for element in elements.elements {
```
brayniac, line 1914: This feels a bit silly, but I think that the only better option is to have the `ValuesWithScores` struct either not exist, or implement the appropriate iterator traits. Alternatively, we can wrap things up with a new type in this crate to provide the iterator.

src/response/cache_sorted_set_fetch_response.rs
cprice404: @brayniac - FYI Kenny will be out on vacation for another week or so.

This PR may be relevant to your question:

https://github.com/momentohq/client-sdk-rust/pull/142

Kenny and I had started to have a conversation about how we will want to tweak the Rust SDK API on the way to 1.0, to make it more consistent with our other SDKs. He filed that PR to take a few of the APIs in the direction that made sense to him after our convo, and I think that our goal is for new APIs to look more like those (as well as going back and migrating the rest of the existing APIs to that pattern when we can, prior to 1.0).

I don't know if that answers your question? I have been meaning to spend some time in this repo to try to get ramped up on the delta between the "old" and "new" patterns that Kenny was trying to lay out, but I haven't had time yet so I don't know if I can give a great review.

That said, I'll try to budget some time for it later this week and see if I can grok enough to weigh in so that you don't have to wait all the way until Kenny returns.

src/simple_cache_client.rs
cprice404: not sure the cleanest way to map it to rust, but in the other SDKs, if you omit "start" it means "everything from the first index to the specified 'end'". If you omit "end" it means "everything from the specified 'start' to the last item in the set." That latter one in particular is important because the caller can't know for sure how many items are in the set when they make the call.

let me know if that answers your question?

src/simple_cache_client.rs
cprice404: yes, our intent is to provide a concrete Response object type for all APIs, as with Kenny's PR that I linked earlier.

Here's what this response type looks like in the Java SDK for example: https://github.com/momentohq/client-sdk-java/blob/main/momento-sdk/src/main/java/momento/sdk/responses/cache/sortedset/SortedSetFetchResponse.java#L23

not sure if that helps?

src/response/cache_sorted_set_fetch_response.rs
brayniac: Thanks! That's helpful.

src/simple_cache_client.rs
brayniac: Yeah - so the question is more about the meaning in momento-protos. In Rust, the range type allows both start and end to be inclusive/exclusive/unbounded. It appears the momento sorted_set_fetch_request::ByIndex allows start to be Some(Inclusive) Some(Unbounded) or None. It's this duplication of the Some(Unbounded) and None that is a bit confusing in the underlying protos library. We're hiding it from the users of this crate, but it's unclear if Some(Unbounded) or None is a preferred representation when building the request.

src/simple_cache_client.rs
brayniac: @cprice404 - it's actually a similar issue here. We have Found(Option<Vec<_>>) or Missing. It's unclear to me if there's some meaningful difference between the protocol crate response being a Missing or a Found(None) or a Found(empty vec). Again, we can hide this from the users, but I'm finding it hard to use the protocol crate to add this functionality to the client given the ambiguities.

I'm personally not finding the Node or Java SDKs to be super helpful as references.

brayniac:
@cprice404 - took another pass which should have it feeling more like the reworked parts of the sdk. Please take a look when you can. Once this is good and merged I'll follow-up with another PR for sorted_set_fetch_by_score.

src/simple_cache_client.rs
cprice404: got it, thanks for the clarification. In the other SDKs we are sending `Unbounded`:

https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/data-client.ts#L2133

src/simple_cache_client.rs
cprice404: so, the response type from this should end up being either `Hit` or `Miss` (like Kenny's DictionaryFetch implementation).

`Miss` would mean that the collection does not exist (and I think corresponds to `Missing` in the protos if I am reading your comment correctly).

`Hit` means that the collection exists, but the vector could be empty.

I think that the reason that VluesWithScores vec could be optional in the proto is because we did `oneof _ValuesWithScores / _Values`, corresponding to the `with_scores` boolean on the request object.

I think that was done on the server side before we discussed the public / SDK API, and we have since decided that we do not want/need to support that `with_scores` option. Thus, in the implementation, with_scores can be hard-coded to true, and we would always expect the Option for ValuesWithScores to be populated. If it is not, that would be considered a bug/error.

Does that help?

src/response/cache_sorted_set_fetch_response.rs
```diff
@@ -1,7 +1,159 @@
...
-    pub value: Option<Vec<SortedSetElement>>,
+    pub value: Option<SortedSet>,
+}
+
+pub enum SortedSetFetch {
+    Found { elements: Elements },
+    Missing,
```
cprice404, line 14: In our public API these are always called `Hit` and `Miss`.

src/response/cache_sorted_set_fetch_response.rs
```diff
@@ -1,7 +1,159 @@
...
+use crate::MomentoError;
+use core::convert::TryFrom;
+use momento_protos::cache_client::sorted_set_fetch_response::SortedSet;
 
 #[derive(Debug)]
 #[non_exhaustive]
 pub struct MomentoSortedSetFetchResponse {
-    pub value: Option<Vec<SortedSetElement>>,
+    pub value: Option<SortedSet>,
+}
```
cprice404, line 10: apologies for any confusion / learning curve on my part as I am coming up to speed on this repo. But looking at Kenny's implementation for DictionaryFetch I think you're right, this struct goes away in favor of the enum below?

src/response/cache_sorted_set_fetch_response.rs
```diff
@@ -1,7 +1,159 @@
...
+                }),
+            },
+            SortedSetFetch::Missing => Err(MomentoError::Miss {
+                description: std::borrow::Cow::Borrowed("sorted set was not found"),
+            }),
+        }
+    }
+}
+
+impl TryFrom<SortedSetFetch> for Vec<Vec<u8>> {
```
cprice404, line 48: If I'm reading this right, this would correspond to `with_scores=false` on the request, which we don't support for now, so you can probably get rid of this?

src/response/cache_sorted_set_fetch_response.rs
```diff
@@ -1,7 +1,159 @@
...
+                }),
+            },
+            SortedSetFetch::Missing => Err(MomentoError::Miss {
+                description: std::borrow::Cow::Borrowed("sorted set was not found"),
+            }),
+        }
+    }
+}
+
+impl TryFrom<SortedSetFetch> for Vec<String> {
```
cprice404, line 109: same here, I think we can get rid of this for now

src/simple_cache_client.rs
```diff
@@ -1873,6 +1873,132 @@ impl SimpleCacheClient {
...
+    /// # })
+    /// # }
+    /// ```
+    pub async fn sorted_set_fetch_by_index(
+        &mut self,
+        cache_name: &str,
+        set_name: impl IntoBytes,
+        order: sorted_set::Order,
+        range: impl RangeBounds<i32>,
+        with_scores: bool,
```
cprice404, line 1931: we can remove this from the public API for now and hard-code to true

src/simple_cache_client.rs
```diff
@@ -1873,6 +1873,132 @@ impl SimpleCacheClient {
...
+    ///
+    /// // this will fetch the first 10 elements (with their scores) in the
+    /// // sorted set, sorted by index (rank) in ascending order. Any valid
+    /// // `core::ops::Range` can be used to select and limit the returned
+    /// // elements.
+    /// let result = momento.sorted_set_fetch_by_index(
+    ///     &cache_name,
+    ///     "test sorted set",
+    ///     Order::Ascending,
+    ///     0..10,
```
cprice404, line 1910: that's cool :)

src/simple_cache_client.rs
cprice404: but if it behaves the way you expect with explicit `Some`, I'm not aware of a downside. At the end of the day all we care about is that we can write integration tests for all of the expected behaviors and they pass :)

src/simple_cache_client.rs
```diff
@@ -1873,6 +1873,132 @@ impl SimpleCacheClient {
...
+            .into_inner();
+
+        // this flattens the response returning a None for both a missing sorted
+        // set and for a request that returns Found with elements being None and
+        // converting the SortedSet enum into the interior collection of
+        // elements.
+        match response.sorted_set {
+            Some(crate::sorted_set::SortedSet::Found(elements)) => match elements.elements {
+                Some(elements) => Ok(SortedSetFetch::Found { elements }),
+                None => Ok(SortedSetFetch::Missing),
```
cprice404, line 1993: I agree with you on the flattening but I think we would just return a `Hit` with an empty vec in this case.

cprice404 (commented):
thanks @brayniac ! I think this looks pretty much right, just a handful of nits and attempts to respond to your questions!

src/simple_cache_client.rs
brayniac: That simplifies things quite a bit. Will do

src/response/cache_sorted_set_fetch_response.rs
brayniac: Correct - I should revert this. We'll leave this type in-place until the old `sorted_set_fetch` function is removed

src/simple_cache_client.rs
brayniac: Updated the example / doctest and it works fine.

brayniac:
Tests updated and added better error handling if we get a response that is values without scores. Should be ready for final review.

src/response/cache_sorted_set_fetch_response.rs
cprice404: aha, I see. Makes sense.

cprice404 (approved):
looks good, thanks Brian!

PR #149, opened 2023-07-11
feat: add sorted_set_fetch_by_score

brayniac:
Adds a simple form of sorted_set_fetch_by_score that allows fetching a range of elements from a sorted set, selecting by score.

### DIFF
--- a/src/simple_cache_client.rs
+++ b/src/simple_cache_client.rs
@@ -1881,10 +1881,10 @@ impl SimpleCacheClient {
     ///
     /// # Arguments
     ///
-    /// * `_cache_name` - name of cache.
-    /// * `_set_name` - name of the set.
-    /// * `_order` - specify ascending or descending order
-    /// * `_range` - constrain to a range of elements by index or by score
+    /// * `cache_name` - name of cache.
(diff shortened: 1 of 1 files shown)


src/simple_cache_client.rs
```diff
@@ -2050,6 +2050,191 @@ impl SimpleCacheClient {
...
+        let min = match range.start_bound() {
+            Bound::Included(v) => Min::MinScore(Score {
+                score: *v,
+                exclusive: false,
+            }),
+            Bound::Excluded(v) => Min::MinScore(Score {
+                score: *v,
+                exclusive: true,
+            }),
+            Bound::Unbounded => Min::UnboundedMin(Unbounded {}),
```
cprice404, line 2169: following up on a question I asked in slack that I think I figured out the answer to; since I'm not very familiar with rust's Range type, I'm guessing that if the user wanted to express "fetch the last 5 items", and we were trying to get a start index of "undefined" and an end index of -5, they could express that via a Rust range of `..-5`?

cprice404 (approved):
lgtm assuming my interpretation of the unbounded range thing is accurate

PR #154, opened 2023-10-05
chore: update protos

kvcache:
Also removes generate_api_token temporarily, as it is currently not
supported in the way this sdk has it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,26 +18,26 @@ members = [
 exclude = [ "example" ]
 
 [dependencies]
-momento-protos = { version = "0.49.1" }
-log = "0.4.17"
+momento-protos = { version = "0.84.1" }
+log = "0.4"
 hyper = { version = "0.14" }
 h2 = { version = "0.3" }
-tonic = { version = "0.8", features = ["tls", "tls-roots", "tls-webpki-roots"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -18,26 +18,26 @@ members = [
...
-uuid = { version = "1.2.2", features = ["v4"] }
-futures = "0.3.25"
-anyhow = "1.0.68"
+base64-url = "2"
+env_logger = "0"
+tokio = { version = "1", features = ["full"] }
+tokio-test = "0"
+uuid = { version = "1", features = ["v4"] }
+futures = "0"
+anyhow = "1"
```
malandis, line 42: Did you mean to remove the patch and minor versions across this? And should the application dependencies be pinned vs lower bounded with semver? eg `log = "0.4"` vs `log = "^0.4"`?

kvcache: For dev dependencies I don't think it matters. It's easier for the next person to `cargo update` if major versions aren't changing.

malandis: OK. I'll open a ticket to unpin the application dependencies.

malandis (approved):

Cargo.toml
kvcache: Cool, thank you - that's worth considering!

PR #172, opened 2024-02-16
chore: disable releasing on every merge to main

nand4011:
Disable releasing on merge to main. We are putting substantial work into the SDK and don't need to create a ton of new releases. We will add release-please at a later time.


bruuuuuuuce (approved):
:shipit:

PR #173, opened 2024-02-21
chore: remove next token from list_caches and list_signing_keys

nand4011:
Remove the next_token argument from list_caches and list_signing_keys. It isn't used by the back end and clutters the function signature.

Add a delete_cache to an integration test that was missing one. The tests may still leak caches if the tests fail before getting to the delete cache statement at the end. More comprehensive protection will follow.


malandis (commented):
LGTM just one question about protobuf defaults

cprice404 (approved):

PR #175, opened 2024-02-26
feat: prototype of new API using requests as builders

cprice404:
This is a prototype of the new API that Kenny, Nate and I have been discussing. In this model we define a Request and a Response struct for each API. The Request structs can be used as builders, to support optional arguments and future additions to the API. The CacheClient exposes the simplest possible syntax for each API, with no optional arguments provide, and implements that by instantiating one of the Request structs.

The Request trait requires each Request type to implement a `send` function, to do the work of sending the request. This will allow us to keep the logic for the implementation details of each API in its own struct, rather than having all of the implementations live direction in `CacheClient`.

`CacheClient` also provides a `send_request` fn that can be called directly, accepting a `Request` type as an argument. This is the lower-level API that callers will use if they need to specify any of the optional arguments for any given request.

### DIFF
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -1,6 +1,6 @@
 use crate::grpc::header_interceptor::HeaderInterceptor;
 use crate::requests::cache::set_add_elements::{SetAddElements, SetAddElementsRequest};
-use crate::requests::cache::{MomentoRequest, MomentoSendableRequest};
+use crate::requests::cache::MomentoRequest;
 use crate::utils::user_agent;
 use crate::{utils, CredentialProvider, IntoBytes, MomentoResult};
 use momento_protos::cache_client::scs_client::ScsClient;
@@ -89,10 +89,7 @@ impl CacheClient {
     /// #
     /// }
     /// ```
-    pub async fn send_request<R: MomentoRequest>(self, request: R) -> MomentoResult<R::Response>
-    where
-        R: MomentoSendableRequest<R>,
-    {
+    pub async fn send_request<R: MomentoRequest>(self, request: R) -> MomentoResult<R::Response> {
         request.send(&self).await
     }
 
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -89,10 +89,7 @@ impl CacheClient {
     /// #
     /// }
     /// ```
-    pub async fn send_request<R: MomentoRequest>(
-        self,
-        request: R,
-    ) -> MomentoResult<R::Response>
+    pub async fn send_request<R: MomentoRequest>(self, request: R) -> MomentoResult<R::Response>
     where
         R: MomentoSendableRequest<R>,
     {
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -49,7 +49,7 @@ impl CacheClient {
     /// let cache_client = momento::CacheClient::new(credential_provider, Duration::from_secs(5))?;
     ///
     /// let set_add_elements_response = cache_client.set_add_elements(cache_name.to_string(), "set", vec!["element1", "element2"]).await?;
-    /// assert_eq!(set_add_elements_response, SetAddElements::Success {});
+    /// assert_eq!(set_add_elements_response, SetAddElements {});
     /// # Ok(())
     /// # })
     /// #
@@ -83,7 +83,7 @@ impl CacheClient {
     /// let set_add_elements_response = cache_client.send_request(
     ///     SetAddElementsRequest::new(cache_name.to_string(), "set", vec!["element1", "element2"])
     /// ).await?;
-    /// assert_eq!(set_add_elements_response, SetAddElements::Success {});
+    /// assert_eq!(set_add_elements_response, SetAddElements {});
     /// # Ok(())
     /// # })
     /// #
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -1,13 +1,13 @@
 use crate::grpc::header_interceptor::HeaderInterceptor;
 use crate::requests::cache::set_add_elements::{SetAddElements, SetAddElementsRequest};
 use crate::requests::cache::{MomentoRequest, MomentoSendableRequest};
-use crate::utils::{user_agent};
+use crate::utils::user_agent;
 use crate::{utils, CredentialProvider, IntoBytes, MomentoResult};
 use momento_protos::cache_client::scs_client::ScsClient;
-use std::convert::{TryInto};
+use std::convert::TryInto;
 use std::time::Duration;
 use tonic::codegen::InterceptedService;
-use tonic::transport::{Channel};
+use tonic::transport::Channel;
 
 pub struct CacheClient {
     pub(crate) data_client: ScsClient<InterceptedService<Channel, HeaderInterceptor>>,
@@ -65,7 +65,6 @@ impl CacheClient {
         request.send(self).await
     }
 
-
     /// ```
     /// # fn main() -> anyhow::Result<()> {
     /// # use momento_protos::cache_client::update_ttl_response::Result::Set;
@@ -90,12 +89,12 @@ impl CacheClient {
     /// #
     /// }
     /// ```
-    pub async fn send_request<R: MomentoRequest>
-    (
+    pub async fn send_request<R: MomentoRequest>(
         self: &Self,
         request: R,
     ) -> MomentoResult<R::Response>
-        where R: MomentoSendableRequest<R>
+    where
+        R: MomentoSendableRequest<R>,
     {
         request.send(self).await
     }
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -1,6 +1,6 @@
 use crate::grpc::header_interceptor::HeaderInterceptor;
 use crate::requests::cache::set_add_elements::{SetAddElements, SetAddElementsRequest};
-use crate::requests::cache::{MomentoRequest, MomentoResponse};
+use crate::requests::cache::{MomentoRequest, MomentoSendableRequest};
 use crate::utils::{user_agent};
 use crate::{utils, CredentialProvider, IntoBytes, MomentoResult};
 use momento_protos::cache_client::scs_client::ScsClient;
@@ -62,7 +62,7 @@ impl CacheClient {
         elements: Vec<E>,
     ) -> MomentoResult<SetAddElements> {
         let request = SetAddElementsRequest::new(cache_name, set_name, elements);
-        MomentoRequest::send(request, self).await
+        request.send(self).await
     }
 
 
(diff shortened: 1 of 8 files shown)


cprice404:
Note that since we needed to rename `SimpleCacheClient` to `CacheClient` on our way to 1.0 anyway, I decided that in this PR I would just introduce a new `CacheClient` and start fleshing out the new API there. Once we get aligned on a few APIs there, we can start migrating things over, and just delete `SimpleCacheClient` when we're done. Might make it a little easier to keep track of things that need to be migrated/moved as we go.

src/cache_client.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// }
+    /// ```
+    pub async fn set_add_elements<E: IntoBytes>(
+        self: &Self,
+        cache_name: String,
+        set_name: impl IntoBytes,
+        elements: Vec<E>,
+    ) -> MomentoResult<SetAddElements> {
+        let request = SetAddElementsRequest::new(cache_name, set_name, elements);
+        MomentoRequest::send(request, self).await
```
kvcache, line 65: do you mean `request.send(self)`?

src/requests/cache/mod.rs
```diff
@@ -0,0 +1,9 @@
+use crate::{MomentoResult};
+use crate::cache_client::CacheClient;
+
+pub mod set_add_elements;
+
+pub trait MomentoResponse {}
+pub trait MomentoRequest<R: MomentoResponse> {
+    fn send(self: Self, cache_client: &CacheClient) -> impl std::future::Future<Output = MomentoResult<R>> + Send;
+}
```
kvcache, line 9: `send()` is not the intended way of sending requests with this api right? To help make the api have 1 shape to follow did you consider something like

```suggestion
pub trait MomentoResponse {}
pub trait MomentoRequest {
    type Response: MomentoResponse;
}

trait SendableRequest<R: MomentoRequest> {
  fn send(self: Self, cache_client: &CacheClient) -> impl std::future::Future<Output = R::Response> + Send;
}
```

which lets you publicly advertise request and response type bindings without exposing the send() function to users?

Also, what purpose does `MomentoResponse` serve? Any reason not to just have
```rust
pub trait MomentoRequest {
    type Response;
}
```
?

Addressing this would simplify send_request to
```rust
pub async fn send_request<R: MomentoRequest>(...) -> MomentoResult<R::Response>
```
which is nice because (1) it only has 1 variadic and (2) there is only 1 valid variadic value of response per request kind anyway, and this expresses that constraint succinctly.

src/requests/cache/set_add_elements.rs
```diff
@@ -0,0 +1,82 @@
...
+                    refresh_ttl: collection_ttl.refresh(),
+                },
+            )?;
+
+            let _ = cache_client
+                .data_client
+                .clone()
+                .set_union(request)
+                .await?
+                .into_inner();
```
kvcache, line 71: spurious `into_inner()` bound to `_`. You should remove this line

src/requests/cache/set_add_elements.rs
```diff
@@ -0,0 +1,82 @@
...
+            Ok(SetAddElements::Success {})
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum SetAddElements {
+    Success {},
+}
```
kvcache, line 80: hmm what other possibilities for the response are there than success, given that `send` already returns `MomentoResult` which encapsulates everything other than Success? Shouldn't this be a struct like `pub struct SetAddElements;`?

kvcache (commented):
I like the direction this is headed

src/requests/cache/mod.rs
cprice404: this is the first i've heard/seen of `type`. I'll have to read up on that.

src/requests/cache/set_add_elements.rs
cprice404: lol this is just literal copypaste from the existing impl but I'll go over it again.

src/requests/cache/set_add_elements.rs
cprice404: I modeled this after what you did for `Get` a while back.

Are you saying we want to have `enum` types for ones that we know can have more than one possible response type, and non-enum types for ones that we don't (currently) think will have more than one response type? it's mostly just another symmetry question in terms of how it will feel to consume the sdk.

kvcache: Result is already an enum type. Some apis are modeled in such a way that a partial success is still a success. For those, a success disposition enum makes sense. For the rest, I see no clear reason to pollute their result types with enums. Do you?

cprice404: i don't have any reasons that I can argue are overwhelmingly compelling, but my reasons would be:

1. symmetry between the different APIs in the Rust SDK
2. future-proofing in case we did add new response types in the future
3. consistency with how we have modeled this in the other language SDKs

if none of those are compelling to you then we can try a few the way that you are describing and decide if it feels too asymmetrical or not.

src/cache_client.rs
cprice404: Thanks for catching this, i feel like at some point it wouldn't compile the other way but I don't remember why, and it does now, and your suggestion is obviously better. fixed in https://github.com/momentohq/client-sdk-rust/pull/175/commits/fb50255179e50212ff6ade6de83f0ed55f0da307

src/requests/cache/mod.rs
cprice404: I did not know about `type` - now that I've read up on it / played with it, I like that much better and love that we get rid of a level of generics. I also agree that there's probably no value in `MomentoResponse` now. Those changes are made.

w/rt separating the `send` out into a different trait, in theory I like that idea; in practice, in order for `CacheClient` to (a) have a public `send_request` fn, and (b) that `fn` to require a `SendableRequest` as a trait bound on its argument, it forces the `SendableRequest` to be public anyway (or else the compiler gives a warning), so I'm not 100% sure whether it's worth separating these traits. I agree that `MomentoRequest.send` isn't the expected public API, but I don't necessarily care if curious users end up calling it, so not 100% sure if it is worth splitting out into 2 traits. But I did it in https://github.com/momentohq/client-sdk-rust/pull/175/commits/fb50255179e50212ff6ade6de83f0ed55f0da307 anyway just to see what it looks like.

src/requests/cache/set_add_elements.rs
cprice404: fixed in https://github.com/momentohq/client-sdk-rust/pull/175/commits/fb50255179e50212ff6ade6de83f0ed55f0da307

src/requests/cache/set_add_elements.rs
cprice404: I changed this in https://github.com/momentohq/client-sdk-rust/pull/175/commits/fb50255179e50212ff6ade6de83f0ed55f0da307 . I'm still on the fence about this but willing to try it out for a while if you feel it's the right way to go.

kvcache (commented):
I’m not personally 100% sure about the MomentoRequest trait as it is anymore. Traits enable user extensibility and I think this usage is much closer to an enumeration, as users can’t extend the internal types and client from the public api. Making it a trait leaks functions and allows extension in an unintended way. Maybe consider using an enumeration for request type definition? I’m not certain it’s ideal that way either.

src/requests/cache/mod.rs
```diff
@@ -0,0 +1,15 @@
+use crate::cache_client::CacheClient;
+use crate::MomentoResult;
+
+pub mod set_add_elements;
+
+pub trait MomentoRequest {
+    type Response;
+}
+
+pub trait MomentoSendableRequest<R: MomentoRequest> {
```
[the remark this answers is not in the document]
cprice404: It's `pub` because it appears in the signature of the `public fn send_request` on the CacheClient. The trait was fine with it being private.

src/simple_cache_client.rs
```diff
@@ -31,11 +31,11 @@ use crate::utils;
 use crate::{grpc::header_interceptor::HeaderInterceptor, utils::connect_channel_lazily};
 use crate::{utils::user_agent, MomentoResult};
 
-pub trait IntoBytes {
+pub trait IntoBytes: Send {
```
cprice404, line 34: @kvcache I also meant to check in with you to make sure this is sane. I think it is, but wanted to doublecheck.

src/requests/cache/mod.rs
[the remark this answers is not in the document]
kvcache: Yeah, I was kind of saying this but pointing out that the trait's orientation is the root of the issue.

I think it's best to remove the duplicate trait because you can't remove the trait from the public signature. I didn't realize that yesterday - by making this function with a MomentoRequest trait public, you permit users to impl MomentoRequest for anything of their own, and for that implementation to be valid, send() would be required per the orientation of the api in this pr.

Since that mechanism is mandated, I'd put the fn back in MomentoRequest
```
/// An internal tool intended to be used by Momento request types. You can impl
/// this fn for your own types if you'd like to hand them to the Momento client
/// directly, but that is not currently an explicitly supported scenario and this trait
/// may change a little over time. If that's okay with you and you want to, impl away!
#[doc(hidden)]
fn send(
    self,
    cache_client: &CacheClient,
) -> impl std::future::Future<Output = MomentoResult<Self::Response>> + Send;
```

src/simple_cache_client.rs
kvcache: Requiring Send on IntoBytes is reasonable - but you could avoid the constraint by eagerly transforming to bytes, before you fork a Send context (before you go async). I think it's okay to require Send for these; the 1 thing I checked to make sure there wouldn't be any obvious problems in the medium term future is https://docs.rs/bytes/1.5.0/bytes/struct.Bytes.html#impl-Send-for-Bytes (it's also implemented for BytesMut). I'm not concerned about this constraint.

src/requests/cache/mod.rs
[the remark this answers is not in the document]
cprice404: I would be fine with this approach and I agree that due to the `pub` there's no point in having two traits.

I experimented with the `enum` path for a while and could not find a direction that seemed promising so I'm bailing on that unless you let me know that you want to discuss it further.

src/simple_cache_client.rs
cprice404: If we stick with the pattern where the trait has (effectively) an `async fn send` on Self, then the only place I can see where I could call `into_bytes` before we find ourselves inside of `async fn send` would be in `::new`. Meaning that the request structs themselves would all end up with `Vec<u8>` for all the things, rather than generic `IntoBytes` fields. I think that would probably be mostly okay but it might be annoying in debugging situations if someone was trying to log a request object that would have otherwise had `String` fields.

I think the UX win of supporting generics -> Strings in those structs is probably worth requiring `: Send` on IntoBytes but lmk if you agree.

kvcache: Debuggers tend to be amenable to printing conveniently-ascii byte sequences as ascii strings in my experience. Also, you don’t have a `std::fmt::Debug` bound on IntoBytes, so nobody is doing the log version of that story today or with the code in this pr anyway ;-) right? I do not really think that debugger pattern matters very much. Removing the bound and storing a plain vec might make some code simpler to read, which is maybe even better than a possible luxury for an attached debugger.

nand4011 (approved):

PR #176, opened 2024-03-01
feat: add a configuration object for the cache client

nand4011:
Add a nested configuration object in the style of the other SDKs.

Add pre-built configurations for different environments.

Plumb the config object into the client, connection, and existing call.

### DIFF
--- a/src/config/configuration.rs
+++ b/src/config/configuration.rs
@@ -1,6 +1,6 @@
-use crate::config::grpc_configuration::GrpcConfiguration;
-use crate::config::transport_strategy::TransportStrategy;
 use std::time::Duration;
+
+use crate::config::transport_strategy::TransportStrategy;
 
 /// Configuration for a Momento cache client.
 ///
@@ -9,25 +9,21 @@ use std::time::Duration;
 /// use momento::config::configurations;
 ///
 /// /// Use laptop for local development
-/// let developer_config = configurations::laptop::LATEST;
+/// let developer_config = configurations::laptop::latest();
 /// /// Use in_region for a typical server environment
-/// let server_config = configurations::in_region::V1;
+/// let server_config = configurations::in_region::v1();
 /// ```
 /// If you have specific requirements, configurations can also be constructed manually:
 /// ```
 /// use std::time::Duration;
 /// use momento::config::configuration::Configuration;
-/// use momento::config::grpc_configuration::GrpcConfiguration;
+/// use momento::config::grpc_configuration::GrpcConfigurationBuilder;
 /// use momento::config::transport_strategy::TransportStrategy;
 ///
 /// let config = Configuration {
 ///             transport_strategy: TransportStrategy {
-///                 grpc_configuration: GrpcConfiguration {
-///                     deadline_millis: Duration::from_millis(1000),
-///                     keep_alive_while_idle: true,
-///                     keep_alive_interval: Duration::from_secs(5000),
-///                     keep_alive_timeout: Duration::from_secs(1000),
-///                 },
+///                 grpc_configuration: GrpcConfigurationBuilder::new(Duration::from_millis(1000))
+///                     .build(),
 ///             },
 ///         };
 #[derive(Clone)]
@@ -37,26 +33,12 @@ pub struct Configuration {
 }
 
 impl Configuration {
-    pub fn new(
-        deadline_millis: Duration,
-        keep_alive_while_idle: bool,
-        keep_alive_interval: Duration,
-        keep_alive_timeout: Duration,
-    ) -> Self {
-        Configuration {
-            transport_strategy: TransportStrategy {
-                grpc_configuration: GrpcConfiguration {
-                    deadline_millis,
-                    keep_alive_while_idle,
-                    keep_alive_interval,
-                    keep_alive_timeout,
-                },
-            },
-        }
+    pub fn new(transport_strategy: TransportStrategy) -> Self {
+        Configuration { transport_strategy }
     }
 
     /// Returns the duration the client will wait before terminating an RPC with a DeadlineExceeded error.
     pub fn deadline_millis(&self) -> Duration {
-        self.transport_strategy.grpc_configuration.deadline_millis
+        self.transport_strategy.grpc_configuration.deadline
     }
 }
--- a/src/config/configuration.rs
+++ b/src/config/configuration.rs
@@ -1,0 +1,62 @@
+use crate::config::grpc_configuration::GrpcConfiguration;
+use crate::config::transport_strategy::TransportStrategy;
+use std::time::Duration;
+
+/// Configuration for a Momento cache client.
+///
+/// Static, versioned configurations are provided for different environments:
+/// ```
+/// use momento::config::configurations;
+///
+/// /// Use laptop for local development
+/// let developer_config = configurations::laptop::LATEST;
+/// /// Use in_region for a typical server environment
+/// let server_config = configurations::in_region::V1;
+/// ```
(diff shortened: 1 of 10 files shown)


src/config/configuration.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::time::Duration;
+
+/// Configuration for a Momento cache client.
+///
+/// Static, versioned configurations are provided for different environments:
+/// ```
+/// use momento::config::configurations;
+///
+/// /// Use laptop for local development
+/// let developer_config = configurations::laptop::LATEST;
```
nand4011, line 12: This currently uses the standard naming scheme we have, with a file called `Configuration` representing the cache client configuration and `Configurations` holding the pre-built configs. The other clients would have have prefixed config names like `TopicConfiguration`.
The full path of a pre-built in this scheme would be `momento::config::configurations::laptop::LATEST` and a hypothetical topic config would be `momento::config::topic-configurations::laptop::LATEST`. That feels pretty clunky to me. We could get combine `Configuration` and `Configurations` and rename it after the type of client it is configuring for something like this: `momento::config::cache::laptop::LATEST` and `momento::config::topic::laptop::LATEST`.
I'm open to any naming scheme.

src/simple_cache_client.rs
```diff
@@ -2714,6 +2714,19 @@ pub(crate) fn prep_request<R>(cache_name: &str, request: R) -> MomentoResult<ton
     Ok(request)
 }
 
+pub(crate) fn prep_request_with_timeout<R>(
```
nand4011, line 2717: This obviously shouldn't be here, but I figured we could find a better home for this logic when we remove the old client.

src/utils.rs
```diff
@@ -80,6 +81,19 @@ pub(crate) fn connect_channel_lazily(uri_string: &str) -> Result<Channel, Channe
     Ok(endpoint.connect_lazy())
 }
 
+pub(crate) fn connect_channel_lazily_configurable(
```
nand4011, line 84: Having a general utils class feels strange to me as well. I think we should eventually make a dedicated place for internal grpc logic that isn't client-specific to live.

src/config/configuration.rs
cprice404: keep in mind that we can use the mod.rs files to `pub use` stuff and make the namespaces look however we want. I think exporting these as `momento::configurations::laptop::latest()` is probably okay but we can let kenny or other rust folks weigh in at some point. it's fungible up until we do a 1.0 so I don't think we have to bottom out on it now.

I'm a little thrown by the `LATEST` (const) without having looked at the PR yet, I would think it'd still need to be a function as opposed to a const, but maybe I'm wrong. I think in the other SDKs mostly we make it a function when we need to support passing in a logging config, and maybe that's not an issue in rust, maybe logging is configured statically like in Java and some other languages, I don't remember :)

src/config/configuration.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Low-level options for network interactions with Momento.
+    pub transport_strategy: TransportStrategy,
+}
+
+impl Configuration {
+    pub fn new(
+        deadline_millis: Duration,
+        keep_alive_while_idle: bool,
+        keep_alive_interval: Duration,
+        keep_alive_timeout: Duration,
```
cprice404, line 44: this definitely needs to be a builder pattern. Most of these will have defaults, and we will definitely end up adding new options over time. Also I don't think we should expose the keepalive settings directly at this level since they are a low-level concern related to the transport strategy.

src/config/configurations.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// V1 config for a typical lambda environment.
+    ///
+    /// This config is guaranteed not to change in future releases of the Momento Rust SDK.
+    pub const V1: Configuration = Configuration {
+        transport_strategy: TransportStrategy {
+            grpc_configuration: GrpcConfiguration {
+                deadline_millis: Duration::from_millis(1100),
+                keep_alive_while_idle: false,
+                keep_alive_interval: Duration::MAX,
+                keep_alive_timeout: Duration::MAX,
```
cprice404, line 112: we need to be really careful with these. I feel pretty confident that the equivalent of `null` or `undefined` for these keepalive settings means something very different than `Duration::MAX`. And I think we want the former here, not the latter.

nand4011: I'll do some testing. I based it off of the .NET sdk, where we use max duration, but assuming that it will behave the same across languages may not be the best plan.

src/config/configuration.rs
nand4011: I updated them to functions instead of constants.

cprice404: cool. Now that i think about it, since rust doesn't support default args or function overloads, if these sigs ever need to change then we will have to create more functions anyway, so it doesn't matter that much. but at least then they'll all be functions, so i think it's good to have changed it.

src/simple_cache_client.rs
cprice404: agreed

src/config/configuration.rs
```diff
@@ -1,33 +1,29 @@
...
 ///
 /// let config = Configuration {
 ///             transport_strategy: TransportStrategy {
-///                 grpc_configuration: GrpcConfiguration {
-///                     deadline_millis: Duration::from_millis(1000),
-///                     keep_alive_while_idle: true,
-///                     keep_alive_interval: Duration::from_secs(5000),
-///                     keep_alive_timeout: Duration::from_secs(1000),
-///                 },
+///                 grpc_configuration: GrpcConfigurationBuilder::new(Duration::from_millis(1000))
```
cprice404, line 25: can we make this `GrpcConfiguration::builder(duration)`?

src/config/configuration.rs
```diff
@@ -1,33 +1,29 @@
...
 /// ```
 /// If you have specific requirements, configurations can also be constructed manually:
 /// ```
 /// use std::time::Duration;
 /// use momento::config::configuration::Configuration;
-/// use momento::config::grpc_configuration::GrpcConfiguration;
+/// use momento::config::grpc_configuration::GrpcConfigurationBuilder;
 /// use momento::config::transport_strategy::TransportStrategy;
 ///
 /// let config = Configuration {
```
cprice404, line 23: I think that if we're going to show an example of this here, we need to do it via `Configuration::builder().blah` as well

src/config/configuration.rs
```diff
@@ -1,33 +1,29 @@
...
 ///             transport_strategy: TransportStrategy {
```
cprice404, line 24: and here; `TransportStrategy::builder()`

src/config/grpc_configuration.rs
```diff
@@ -21,3 +21,67 @@ pub struct GrpcConfiguration {
     /// closing the connection.
     pub keep_alive_timeout: Duration,
 }
+
+impl GrpcConfiguration {
+    pub fn new(
```
cprice404, line 26: I don't think we should expose this?

src/config/transport_strategy.rs
```diff
@@ -6,3 +6,9 @@ pub struct TransportStrategy {
     /// Low-level gRPC settings for communicating with Momento.
     pub grpc_configuration: GrpcConfiguration,
 }
+
+impl TransportStrategy {
+    pub fn new(grpc_configuration: GrpcConfiguration) -> Self {
```
cprice404, line 11: same here, i'm thinking we don't want any `pub fn new` on any of these probably?

cprice404 (approved):
I think this looks good for now!

PR #178, opened 2024-03-11
feat: Add sorted set put and fetch functions

nand4011:
Add sorted set put element(s), fetch by rank, and fetch by score.

Add create and delete cache methods for testing.

Add test setup and teardown scripts so that we can guarantee the cache exists for doc and integration tests, and ensure it is deleted after they run.

Call the setup and teardown scripts in github ci.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,6 +51,7 @@ jobs:
       - name: Cargo Check
         run: cargo check
       - name: Test Setup
+        # This script relies on dev dependencies such as Tokio, so we run it with the --example flag
         run: cargo run --example test-setup
       - name: Unit Tests
         run: cargo test --lib
@@ -59,7 +60,9 @@ jobs:
       - name: Integration Tests
         run: cargo test --tests
       - name: Test Teardown
+        # We want the teardown to execute even if an earlier step fails, hence the `always()` condition
         if: always()
+        # This script relies on dev dependencies such as Tokio, so we run it with the --example flag
         run: cargo run --example test-teardown
 
   rustfmt-build-examples:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,7 +34,6 @@ jobs:
     runs-on: macos-latest
     env:
       MOMENTO_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
-      TEST_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
       # TODO: get rid of this once everything has been migrated to MOMENTO_API_KEY
       TEST_AUTH_TOKEN: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,6 +7,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  TEST_CACHE_NAME: rust-integration-test-ci-${{ github.sha }}
 
 jobs:
   rustfmt:
@@ -33,6 +34,7 @@ jobs:
     runs-on: macos-latest
     env:
       MOMENTO_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
+      TEST_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
       # TODO: get rid of this once everything has been migrated to MOMENTO_API_KEY
       TEST_AUTH_TOKEN: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
 
@@ -48,12 +50,17 @@ jobs:
         run: cargo build --verbose
       - name: Cargo Check
         run: cargo check
+      - name: Test Setup
+        run: cargo run --example test-setup
       - name: Unit Tests
         run: cargo test --lib
       - name: Doc Tests
         run: cargo test --doc
       - name: Integration Tests
         run: cargo test --tests
+      - name: Test Teardown
+        if: always()
+        run: cargo run --example test-teardown
 
   rustfmt-build-examples:
     name: Style & Lint
--- a/scripts/test-setup.rs
+++ b/scripts/test-setup.rs
@@ -1,8 +1,6 @@
 use std::env;
 use std::time::Duration;
 
-use tokio;
-
 use momento::config::configurations;
 use momento::{CacheClient, CredentialProviderBuilder};
 
--- a/scripts/test-setup.rs
+++ b/scripts/test-setup.rs
@@ -1,8 +1,7 @@
-use std::env;
 use std::time::Duration;
 
 use momento::config::configurations;
-use momento::{CacheClient, CredentialProviderBuilder};
+use momento::{CacheClient};
 use momento_test_util::{get_test_cache_name, get_test_credential_provider};
 
 #[tokio::main]
--- a/scripts/test-setup.rs
+++ b/scripts/test-setup.rs
@@ -1,7 +1,7 @@
 use std::time::Duration;
 
 use momento::config::configurations;
-use momento::{CacheClient};
+use momento::CacheClient;
 use momento_test_util::{get_test_cache_name, get_test_credential_provider};
 
 #[tokio::main]
--- a/scripts/test-setup.rs
+++ b/scripts/test-setup.rs
@@ -1,0 +1,31 @@
+use std::env;
+use std::time::Duration;
+
+use tokio;
+
+use momento::config::configurations;
+use momento::{CacheClient, CredentialProviderBuilder};
+
+#[tokio::main]
+async fn main() {
+    let cache_name =
+        env::var("TEST_CACHE_NAME").expect("environment variable TEST_CACHE_NAME should be set");
+
+    let credential_provider =
+        CredentialProviderBuilder::from_environment_variable("TEST_API_KEY".to_string())
+            .build()
+            .expect("auth token should be valid");
+
+    let cache_client = CacheClient::new(
+        credential_provider,
+        configurations::laptop::latest(),
+        Duration::from_secs(5),
+    )
+    .expect("cache client cannot be created");
+
+    println!("Creating cache {}", cache_name.clone());
+    cache_client
+        .create_cache(cache_name.clone())
+        .await
+        .expect("cache cannot be created");
+}
--- a/scripts/test-setup.rs
+++ b/scripts/test-setup.rs
@@ -3,16 +3,13 @@ use std::time::Duration;
 
 use momento::config::configurations;
 use momento::{CacheClient, CredentialProviderBuilder};
+use momento_test_util::{get_test_cache_name, get_test_credential_provider};
 
 #[tokio::main]
 async fn main() {
-    let cache_name =
-        env::var("TEST_CACHE_NAME").expect("environment variable TEST_CACHE_NAME should be set");
+    let cache_name = get_test_cache_name();
 
-    let credential_provider =
-        CredentialProviderBuilder::from_environment_variable("TEST_API_KEY".to_string())
-            .build()
-            .expect("auth token should be valid");
+    let credential_provider = get_test_credential_provider();
 
     let cache_client = CacheClient::new(
         credential_provider,
--- a/src/requests/cache/create_cache.rs
+++ b/src/requests/cache/create_cache.rs
@@ -4,6 +4,30 @@ use tonic::Request;
 use crate::requests::cache::MomentoRequest;
 use crate::{utils, CacheClient, MomentoResult};
 
+/// Request to create a cache.
+///
+/// # Arguments
+///
+/// * `cache_name` - The name of the cache to create.
+///
+/// # Example
+/// Assumes that a CacheClient named `cache_client` has been created and is available.
+/// ```no_run
+/// # fn main() -> anyhow::Result<()> {
+/// # use momento_test_util::create_doctest_client;
+/// # tokio_test::block_on(async {
+/// use momento::requests::cache::create_cache::CreateCache;
+/// use momento::requests::cache::create_cache::CreateCacheRequest;
+/// # let (cache_client, cache_name) = create_doctest_client();
+///
+/// let create_cache_request = CreateCacheRequest::new(cache_name.to_string());
+///
+/// let create_cache_response = cache_client.send_request(create_cache_request).await?;
+///
+/// assert_eq!(create_cache_response, CreateCache {});
+/// # Ok(())
+/// # })
+/// # }
 pub struct CreateCacheRequest {
     pub cache_name: String,
 }
@@ -34,4 +58,5 @@ impl MomentoRequest for CreateCacheRequest {
     }
 }
 
+#[derive(Debug, PartialEq, Eq)]
 pub struct CreateCache {}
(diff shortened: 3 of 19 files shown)


src/requests/cache/create_cache.rs
```diff
@@ -0,0 +1,62 @@
...
+/// Assumes that a CacheClient named `cache_client` has been created and is available.
+/// ```no_run
+/// # fn main() -> anyhow::Result<()> {
+/// # use momento_test_util::create_doctest_client;
+/// # tokio_test::block_on(async {
+/// use momento::requests::cache::create_cache::CreateCache;
+/// use momento::requests::cache::create_cache::CreateCacheRequest;
+/// # let (cache_client, cache_name) = create_doctest_client();
+///
+/// let create_cache_request = CreateCacheRequest::new(cache_name.to_string());
```
nand4011, line 23: Having to call to_string or when giving strings to these methods feels strange. I tried changing them to &str, but then I need to define a lifetime for the request struct. What is the idiomatic thing to do here?

scripts/test-setup.rs
```diff
@@ -0,0 +1,29 @@
+use std::env;
+use std::time::Duration;
+
+use momento::config::configurations;
+use momento::{CacheClient, CredentialProviderBuilder};
+
+#[tokio::main]
+async fn main() {
+    let cache_name =
+        env::var("TEST_CACHE_NAME").expect("environment variable TEST_CACHE_NAME should be set");
```
cprice404, line 10: any thoughts on providing a default here? it kinda always bugs me for local dev when we have to add another env var, but i can imagine some pros and cons

scripts/test-setup.rs
```diff
@@ -0,0 +1,29 @@
...
+
+    let credential_provider =
+        CredentialProviderBuilder::from_environment_variable("TEST_API_KEY".to_string())
```
cprice404, line 13: supernit: i've been slowly trying to standardize on MOMENTO_API_KEY for these. In the future I can imagine that we might have a default credentialprovider that looks for creds in a few well known places, including a well known env var name, like AWS does. I don't see any super compelling reason to make tests use a different env var name. Thoughts?

src/requests/cache/create_cache.rs
cprice404: I don't think `to_string` is uncommon but I think there might also be a way to specify a signature that can accept both, using some kind of `Into` thing. I'll create a ticket to look into it but we don't need to block this pr on it.

src/requests/cache/set_add_elements.rs
```diff
@@ -42,6 +43,11 @@ impl<S: IntoBytes, E: IntoBytes> SetAddElementsRequest<S, E> {
             collection_ttl: Some(collection_ttl),
         }
     }
+
+    pub fn with_ttl(mut self, collection_ttl: CollectionTtl) -> Self {
```
cprice404, line 47: thanks for adding this.

do you have any strong opinions on how we name these going forward? I was waffling between using the `with_` prefix or not. I might be leaning towards not using that prefix, because in some stuff I've worked with in the past, it was used to signify a copy constructor (thus, leaving the original object intact without mutating it, whereas this is definitely mutating it).

src/requests/cache/sorted_set_fetch_by_rank.rs
```diff
@@ -0,0 +1,143 @@
...
+/// Assumes that a CacheClient named `cache_client` has been created and is available.
+/// ```
+/// # fn main() -> anyhow::Result<()> {
+/// # use std::convert::TryInto;
+/// # use momento_test_util::create_doctest_client;
+/// # tokio_test::block_on(async {
+/// use momento::requests::cache::sorted_set_fetch_by_rank::SortOrder;
+/// use momento::requests::cache::sorted_set_fetch_by_rank::SortedSetFetchByRankRequest;
+/// use momento::response::cache::sorted_set_fetch::SortedSetFetch;
+/// # let (cache_client, cache_name) = create_doctest_client();
```
cprice404, line 43: do you know if this will get excluded from the generated docs, or whether the comment blocks have to be contiguous at the top/bottom of the function in order to get excluded? if you don't know we can just look into this as a follow-up, not a blocker for this PR.

src/requests/cache/sorted_set_fetch_by_rank.rs
```diff
@@ -0,0 +1,143 @@
...
+/// let put_element_response = cache_client.sorted_set_put_elements(
+///     cache_name.to_string(),
+///     sorted_set_name.to_string(),
+///     vec![("value1", 1.0), ("value2", 2.0), ("value3", 3.0), ("value4", 4.0)]
+/// ).await?;
+///
+/// let fetch_request = SortedSetFetchByRankRequest::new(cache_name, sorted_set_name)
+///     .with_order(SortOrder::Ascending)
+///     .with_start_rank(1)
+///     .with_end_rank(3);
```
cprice404, line 55: this looks pretty good to me! same question about `with_` prefix but otherwise I think I like how it looks

src/requests/cache/sorted_set_put_element.rs
```diff
@@ -0,0 +1,105 @@
...
+///     1.0
+/// ).with_ttl(CollectionTtl::default());
+///
+/// let create_cache_response = cache_client.send_request(put_element_request).await?;
+///
+/// assert_eq!(create_cache_response, SortedSetPutElement {});
+/// # Ok(())
+/// # })
+/// # }
+pub struct SortedSetPutElementRequest<S: IntoBytes, E: IntoBytes> {
```
cprice404, line 46: nit I think the convention for naming these generic types is to name them the first letter of whatever field they are going into, so we did `E` for "elements", and this one would be `V` I think.

src/requests/cache/sorted_set_put_elements.rs
```diff
@@ -0,0 +1,105 @@
...
+/// let put_elements_response = cache_client.send_request(put_elements_request).await?;
+///
+/// assert_eq!(put_elements_response, SortedSetPutElements {});
+/// # Ok(())
+/// # })
+/// # }
+pub struct SortedSetPutElementsRequest<S: IntoBytes, E: IntoBytes> {
+    cache_name: String,
+    sorted_set_name: S,
+    elements: Vec<(E, f64)>,
```
cprice404, line 47: I can imagine that we might want to support a Map type as input here, so we might need another Into thing, but we can explore that in a follow-up PR.

test-util/src/lib.rs
```diff
@@ -47,3 +50,24 @@ where
 
     runtime.block_on(func(cache_name, credential_provider))
 }
+
+pub fn create_doctest_client() -> (CacheClient, String) {
+    let cache_name =
+        env::var("TEST_CACHE_NAME").expect("environment variable TEST_CACHE_NAME should be set");
+
+    let credential_provider =
+        CredentialProviderBuilder::from_environment_variable("TEST_API_KEY".to_string())
```
cprice404, line 59: same comments here about whether it would be sane to use a default for test_cache_name if it's missing, and whether we should just use MOMENTO_API_KEY

tests/cache_sorted_set_tests.rs
```diff
@@ -0,0 +1,311 @@
...
+use momento::requests::cache::sorted_set_fetch_by_rank::SortedSetFetchByRankRequest;
+use momento::requests::cache::sorted_set_fetch_by_score::SortedSetFetchByScoreRequest;
+use momento::response::cache::sorted_set_fetch::SortedSetFetch;
+
+use crate::cache_test_state::TEST_STATE;
+
+mod cache_test_state;
+
+#[tokio::test]
+async fn sorted_set_put_element_happy_path() {
```
cprice404, line 13: so, one thing I wanted to sync with you on, it seems like in a lot of rust code the tests are just added to the same file as the actual code that is being tested. idk if we have prior art on that in this repo or if that is sane, or if for us it makes more sense to keep them separate especially since they are integration tests. any thoughts?

tests/cache_test_state.rs
```diff
@@ -0,0 +1,81 @@
...
+
+            sender
+                .send(Some(cache_client))
+                .expect("client should be sent to test state thread");
+            thread_barrier.wait();
+        });
+        barrier.wait();
+
+        // Retrieve the client from the runtime that created it.
+        let client = client_receiver
```
cprice404, line 69: wow, fancy. I understand how this is working but am a little surprised that we have to use a channel to pass it back? naively, it seems like with an Arc we should be able to just reference the same client from multiple threads, but maybe Arcs are just for referencing from multiple functions in the same thread, and channels are required to move across thread boundaries?

anyway, I assume you and kenny paired on this and aligned on this solution, in which case I'm +1, just musing about how it reads to a newb :)

tests/cache_test_state.rs
```diff
@@ -0,0 +1,81 @@
+use std::env;
+use std::sync::Arc;
+use std::time::Duration;
+
+use once_cell::sync::Lazy;
```
cprice404, line 5: did you guys talk about the different crates that are available for this, and just curious if there was a reason this one seemed best?

cprice404 (commented):
this looks great. left a bunch of questions that i'm interested in your feedback on, but no blockers, and i'm happy to pick up any changes that we agree should go into this PR before it gets merged so that you can stay focused on other tasks.

.github/workflows/ci.yaml
```diff
@@ -48,12 +50,17 @@ jobs:
         run: cargo build --verbose
       - name: Cargo Check
         run: cargo check
+      - name: Test Setup
+        run: cargo run --example test-setup
```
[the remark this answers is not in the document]
nand4011: The setup and teardown binaries are defined as examples in Cargo.toml. It is hacky to do that, but normal bins don't have access to dev dependencies, so it would be hard to get access to tokio without locking that dependency behind a feature flag. There might be a better way to organize that, but I couldn't find one.

.github/workflows/ci.yaml
```diff
@@ -48,12 +50,17 @@ jobs:
...
       - name: Unit Tests
         run: cargo test --lib
       - name: Doc Tests
         run: cargo test --doc
       - name: Integration Tests
         run: cargo test --tests
+      - name: Test Teardown
+        if: always()
```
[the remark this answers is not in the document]
nand4011: It will always run even if the job fails or is cancelled. I wanted to make sure that we wouldn't leak caches in those cases. I haven't tested it properly since the tests have always passed, however.

scripts/test-setup.rs
nand4011: That also reminds me that I forgot to make a CONTRIBUTING.md.

scripts/test-setup.rs
nand4011: I'm good with MOMENTO_API_KEY. I agree that there isn't much reason to separate them, other than if someone consistently needed to run a local application using a different key than they were using to develop an sdk.

src/requests/cache/set_add_elements.rs
nand4011: I'm not too attached to it, but I would need to think of another name that felt good being called in a chain, like
```
.with_order(SortOrder::Ascending)
.with_start_rank(1)
.with_end_rank(3);
```
I suppose we could just do
```
.order(SortOrder::Ascending)
.start_rank(1)
.end_rank(3);
```
I also think it would be worth seeing if this style is more ergonomic for building config objects than the explicit builders we're currently using.

src/requests/cache/sorted_set_fetch_by_rank.rs
nand4011: It will exclude any line maked with a hash. Here is what that example looks like:
```
use momento::requests::cache::sorted_set_fetch_by_rank::SortOrder;
use momento::requests::cache::sorted_set_fetch_by_rank::SortedSetFetchByRankRequest;
use momento::response::cache::sorted_set_fetch::SortedSetFetch;
let sorted_set_name = "sorted_set";

let put_element_response = cache_client.sorted_set_put_elements(
    cache_name.to_string(),
    sorted_set_name.to_string(),
    vec![("value1", 1.0), ("value2", 2.0), ("value3", 3.0), ("value4", 4.0)]
).await?;
```

src/requests/cache/sorted_set_put_element.rs
nand4011: Good find. That's a copy/paste error from when I copied it from set_add_elements.

tests/cache_sorted_set_tests.rs
nand4011: I honestly hadn't thought about that. I just saw we had a tests directory with some integration tests in it and my Java instincts kicked in. I would need to do an experiment to see if the global test state would still work the same way, but I'm good with adding them to the same file. My only concern would be if it would become hard to read if the cache client were filled with doctests and integration tests.

tests/cache_test_state.rs
nand4011: Yeah, Kenny and I went through several frustrating steps to get to this point. There may be an easier solution, but I don't currently have the Rust chops to point it out.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
cprice404: nope that seems fine but we might want to put a comment here. i can do it.

scripts/test-setup.rs
cprice404: discussed in slack, the current behavior seems okay, will standardize on a default fallback cache name

scripts/test-setup.rs
cprice404: thx

tests/cache_sorted_set_tests.rs
cprice404: k, will fish around for opinions on the backend team... on this one i could make a case either way.

src/requests/cache/set_add_elements.rs
cprice404: yeah i was kind of thinking of just removing the `with_`, just like you did in your example there.

re: config builders, yeah i will have to dig into that more to form an opinion... in my head it boils down to whether the struct for the builder is the same struct for the final object or not. will keep it in mind though.

src/requests/cache/sorted_set_fetch_by_rank.rs
cprice404: sweet, thanks

tests/cache_test_state.rs
cprice404: if it's in std then why do we need the dep on once_cell? is the std thing only in nightly or something?

tests/cache_test_state.rs
cprice404: ack

src/requests/cache/set_add_elements.rs
cprice404: will create a separate ticket for this

tests/cache_test_state.rs
cprice404: discussed in slack - confirmed that the thing we need here is only in nightly, so we still need the crate

cprice404:
Chris TODO:
- [ ] add comments re: `cargo --example`
- [ ] add comments re: `always()`
- [ ] add default cache name fallback and maybe DRY up?
- [ ] change TEST_API_KEY to MOMENTO_API_KEY and maybe DRY up?
- [ ] change relevant `E` generics to `V`s

cprice404:
@nand4011 I've touched up everything that I was interested in for this PR, and made some tickets for followup items. Would you mind giving it a peek?

cprice404 (approved):

nand4011:
I can't approve it, but it LGTM.

PR #179, opened 2024-03-12
feat: Add sorted set put and fetch functions

cprice404:
Add sorted set put element(s), fetch by rank, and fetch by score.

Add create and delete cache methods for testing.

Add test setup and teardown scripts so that we can guarantee the cache
exists for doc and integration tests, and ensure it is deleted after they run.

Call the setup and teardown scripts in github ci.


cprice404:
This is a dupe of #178 , which I opened because for some reason that PR is not getting updated with new commits when I push them to the branch :/

PR #185, opened 2024-03-18
feat: add get/set to CacheClient to support readme.ts example

cprice404:
The main goal of this PR is to get us closer to having a good example code snippet in the readme.ts file. There were a few things I discovered that we need along the way toward that, so this PR tries to bite off a few of them:

* We need `get` and `set`, so those are added in this PR.

* The credential provider builder pattern didn't seem quite as ergonomic as some of our newer ones (CacheClient/requests), and wasn't consistent with the direction of them, so I tweaked a few things to try to move it in that direction. This is not the final state of this, I intend to do at least one more pass to try to make sure that the builders/constructors for client, requests, cred providers, and configs are as consistent as possible, but this is an interim step.

* I also removed/commented out the support for overriding endpoints in the credential providers for now; these had gotten a little stale relative to the current endpoints support, and I'd rather keep the API surface area as small as possible for now. We will probably have to add these back in in some fashion in order to update the CLI to the new release of the SDK but we can do that when we get there.

I will also add a copy paste of what the example code will look like after this PR, but it can't be merged until we do a release because it exists in the examples dir and must be pinned to a released version of the SDK.

### DIFF
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -8,15 +8,19 @@ use tonic::transport::Channel;
 
 use crate::config::configuration::Configuration;
 use crate::grpc::header_interceptor::HeaderInterceptor;
+use crate::requests::cache::basic::get::{Get, GetRequest};
+use crate::requests::cache::basic::set::{Set, SetRequest};
 use crate::requests::cache::create_cache::{CreateCache, CreateCacheRequest};
 use crate::requests::cache::delete_cache::{DeleteCache, DeleteCacheRequest};
-use crate::requests::cache::set_add_elements::{SetAddElements, SetAddElementsRequest};
-use crate::requests::cache::sorted_set_fetch_by_rank::{SortOrder, SortedSetFetchByRankRequest};
-use crate::requests::cache::sorted_set_fetch_by_score::SortedSetFetchByScoreRequest;
-use crate::requests::cache::sorted_set_put_element::{
+use crate::requests::cache::set::set_add_elements::{SetAddElements, SetAddElementsRequest};
+use crate::requests::cache::sorted_set::sorted_set_fetch_by_rank::{
+    SortOrder, SortedSetFetchByRankRequest,
+};
+use crate::requests::cache::sorted_set::sorted_set_fetch_by_score::SortedSetFetchByScoreRequest;
+use crate::requests::cache::sorted_set::sorted_set_put_element::{
     SortedSetPutElement, SortedSetPutElementRequest,
 };
-use crate::requests::cache::sorted_set_put_elements::{
+use crate::requests::cache::sorted_set::sorted_set_put_elements::{
     SortedSetPutElements, SortedSetPutElementsRequest,
 };
 use crate::requests::cache::MomentoRequest;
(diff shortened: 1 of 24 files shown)


cprice404:
Here's an early example of what the readme.ts example code might look like after these changes:

```rust
use std::time::Duration;
use momento::{CacheClient, CredentialProvider};
use momento::config::configurations::laptop;

const CACHE_NAME: String = "cache";

pub async fn main() {
    let cache_client = CacheClient::new(
        CredentialProvider::from_env_var("MOMENTO_API_KEY")?,
        laptop::latest(),
        Duration::from_secs(60)
    )?;

    cache_client.create_cache(CACHE_NAME).await?;

    match(cache_client.set(CACHE_NAME, "mykey", "myvalue").await) {
        Ok(_) => println!("Successfully stored key 'mykey' with value 'myvalue'"),
        Err(e) => println!("Error: {}", e)
    }
    let item: String = match(cache_client.get(&cache_name, "mykey").await?) {
        Get::Hit { value } => value.try_into().expect("I stored a string!"),
        Get::Miss => return Err(anyhow::Error::msg("cache miss")) // probably you'll do something else here
    };
```

src/credential_provider.rs
```diff
@@ -58,134 +111,102 @@ impl Debug for AuthTokenSource {
 }
 
 #[derive(Debug)]
-pub struct CredentialProviderBuilder {
+struct CredentialProviderBuilder {
```
cprice404, line 114: for now I made this builder private. We use it internally, and some parts of it may become public again in the future, but I wanted to simplify the 90% case for creating a CredentialProvider for now.

src/credential_provider.rs
```diff
@@ -58,134 +111,102 @@ impl Debug for AuthTokenSource {
...
-            &endpoint_override,
-        ));
-        self.control_endpoint_override = Some(CredentialProviderBuilder::get_control_endpoint(
-            &endpoint_override,
-        ));
-        self
-    }
-
-    pub fn build(self) -> MomentoResult<CredentialProvider> {
+    // /// Override the data plane endpoint
```
cprice404, line 137: Here's all the stuff where we were previously allowing endpoint overrides, that I have temporarily commented out. I can delete this stuff instead of commenting it if anyone feels strongly but it felt like it might be useful again in the not-too-distant future.

src/cache_client.rs
```diff
@@ -161,31 +165,182 @@ impl CacheClient {
...
     }
 
+    /// Sets an item in a Momento Cache
+    ///
+    /// # Arguments
+    ///
+    /// * `cache_name` - name of cache
+    /// * `key` - key of the item whose value we are setting
+    /// * `value` - data to stored in the cache item
+    /// * `ttl` - The TTL to use for the
```
nand4011, line 175: Should this be in an optional arguments section?

src/credential_provider.rs
```diff
@@ -31,6 +31,59 @@ pub struct CredentialProvider {
     pub cache_endpoint: String,
 }
 
+impl CredentialProvider {
+    /// Returns a builder to produce a Credential Provider using an API key stored in the specified
```
nand4011, line 35:
```suggestion
    /// Returns a Credential Provider using an API key stored in the specified
```

src/credential_provider.rs
```diff
@@ -31,6 +31,59 @@ pub struct CredentialProvider {
...
+    /// environment variable
+    ///
+    /// # Arguments
+    ///
+    /// * `env_var_name` - Name of the environment variable to read token from
+    /// # Examples
+    ///
+    /// ```
+    /// # tokio_test::block_on(async {
+    ///     use momento::CredentialProvider;
```
nand4011, line 45: Minor thing, but this spacing is preserved in the docs, so the example code will all be indented 4 extra spaces.

src/credential_provider.rs
```diff
@@ -31,6 +31,59 @@ pub struct CredentialProvider {
...
+    ///     let credential_provider = CredentialProvider::from_env_var("MOMENTO_API_KEY".to_string())
+    ///         .expect("MOMENTO_API_KEY must be set");
+    /// # })
+    /// ```
+    ///
+    pub fn from_env_var(env_var_name: String) -> MomentoResult<CredentialProvider> {
+        CredentialProviderBuilder::from_environment_variable(env_var_name).build()
+    }
+
+    /// Returns a builder to produce a Credential Provider from the provided API key
```
nand4011, line 55:
```suggestion
    /// Returns a Credential Provider from the provided API key
```

src/requests/cache/basic/set.rs
```diff
@@ -0,0 +1,68 @@
...
+        self.ttl = Some(ttl);
+        self
+    }
+}
+
+impl<K: IntoBytes, V: IntoBytes> MomentoRequest for SetRequest<K, V> {
+    type Response = Set;
+
+    async fn send(self, cache_client: &CacheClient) -> MomentoResult<Set> {
+        // let ttl = self.ttl.unwrap_or_default();
```
nand4011, line 49: Is this leftover from set add elements?

src/response/cache/sorted_set_fetch.rs
```diff
@@ -6,6 +6,8 @@ use momento_protos::cache_client::SortedSetFetchResponse;
 
 use crate::{MomentoError, MomentoResult};
 
+// TODO this needs to be moved to the requests directory
```
nand4011, line 9: I made it a separate file because it is used by multiple requests. Are we going to define all responses alongside the requests because they are closely tied to them?

src/cache_client.rs
cprice404: oh good catch, will fix, thanks

src/credential_provider.rs
cprice404: ack

src/requests/cache/basic/set.rs
cprice404: oops probably, thanks

src/response/cache/sorted_set_fetch.rs
cprice404: yeah i meant to reach out to you about this and get your opinion.

tbh part of my motivation here is just to help us in tracking milestones toward the release - my current thinking is that the `requests` directory will go away entirely. I'd rather not have to surgically check all the files in there and see which to keep and which to get rid of. So I'm thinking maybe we just move all of the stuff that is going to survive over into the `requests` directory, even though I agree with you that there will be a few files that are shared amongst multiple request types. wdyt?

src/requests/cache/basic/set.rs
cprice404: fixed in https://github.com/momentohq/client-sdk-rust/pull/185/commits/a82609ceedbd3336bc109c3f2b3454a1e1d9dcad

src/cache_client.rs
cprice404: fixed in https://github.com/momentohq/client-sdk-rust/pull/185/commits/a82609ceedbd3336bc109c3f2b3454a1e1d9dcad , ty

cprice404:
@nand4011 I think I fixed everything you found, just the one open question about the plan to move stuff to `requests` dir.

src/response/cache/sorted_set_fetch.rs
nand4011: I made the `cache` directory in `responses` and added the sorted set fetch response there, mostly so we could easily separate them. I am alright moving it though, especially since sorted set fetch is one of the only responses that is reused.

nand4011 (approved):

src/response/cache/sorted_set_fetch.rs
cprice404: maybe we even move it back to responses dir after the old one is deleted, but i think it will be easier to track progress for now if we just plan on getting rid of that whole dir.

PR #187, opened 2024-03-20
chore: set up release please

anitarua:
This is (at the very least) a first attempt at setting up release please in this sdk

### DIFF
--- a/.github/workflows/on-push-to-main.yml
+++ b/.github/workflows/on-push-to-main.yml
@@ -68,6 +68,11 @@ jobs:
     needs: [ release-please ]
     if: ${{ needs.release-please.outputs.release_created == 'true' }}
     steps:
+      - name: Setup repo
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.MOMENTO_MACHINE_USER_GITHUB_TOKEN }}
+          
       - run: |
           cargo publish
           --token ${{ secrets.CARGO_REGISTRY_TOKEN }}
--- a/.github/workflows/on-push-to-main.yml
+++ b/.github/workflows/on-push-to-main.yml
@@ -59,8 +59,6 @@ jobs:
           package-name: momento
           default-branch: main
           changelog-types: '[{"type":"feat","section":"Features","hidden":false},{"type":"fix","section":"Bug Fixes","hidden":false},{"type":"chore","section":"Miscellaneous","hidden":false}]'
-          extra-files: | 
-            Cargo.toml
 
   publish:
     name: Publish to crates.io
--- a/.github/workflows/on-push-to-main.yml
+++ b/.github/workflows/on-push-to-main.yml
@@ -72,8 +72,64 @@ jobs:
         uses: actions/checkout@v3
         with:
           token: ${{ secrets.MOMENTO_MACHINE_USER_GITHUB_TOKEN }}
-          
+
       - run: |
           cargo publish
           --token ${{ secrets.CARGO_REGISTRY_TOKEN }}
     
+
+
+# Previous release process below
+#  release:
+#    runs-on: ubuntu-latest
+#    outputs:
+#      version: ${{ steps.release.outputs.release }}
+#    steps:
+#      - uses: actions/checkout@v3
+#      - name: Set release
+#        id: semrel
+#        uses: go-semantic-release/action@v1
+#        with:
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -1,6 +1,8 @@
 [package]
 name = "momento"
+# x-release-please-start-version
 version = "0.1.0"
+# x-release-please-end
```
anitarua, line 5: Seemed like Rust provides manifest version number as an [env var](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) that's used when sending the `agent` header.
Put these comments here to get release-please to update version number in this file.

.github/workflows/on-push-to-main.yml
```diff
@@ -94,3 +43,32 @@ jobs:
...
+
+  publish:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    needs: [ release-please ]
+    if: ${{ needs.release-please.outputs.release_created == 'true' }}
+    steps:
+      - run: |
+          cargo publish
+          --token ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
anitarua, line 73: Found this [forum post](https://users.rust-lang.org/t/does-anyone-use-github-actions-to-run-cargo-publish/92374/5) that provides token using this `--token` argument instead of doing a `cargo login` step.

Not sure if I need to include `- uses: actions/checkout@v3` as a step before publishing

.github/workflows/on-push-to-main.yml
```diff
@@ -21,61 +25,6 @@ jobs:
...
-#        with:
-#          toolchain: stable
-#          components: rustfmt
-#          override: true
-#
-#      - name: Build
-#        run: cargo build --verbose
-#
-#      - name: Integration Tests
-#        run: cargo test --tests
```
anitarua, line 63: can add this build and test step back into the workflow if we want

anitarua: or I can leave it commented out but bring these lines back into the file if we want to keep them for reference or anything

.github/workflows/on-push-to-main.yml
anitarua: Or I can try to use this github action for publishing crates instead? https://github.com/marketplace/actions/publish-crates

Cargo.toml
cprice404: sorry, not quite following the part about the env var, does that get used here somehow?

.github/workflows/on-push-to-main.yml
cprice404: yeah i think maybe let's keep this stuff in, commented out, until we have our first successful release. Then i'd probably go ahead and add the tests back in to the build steps, but you don't need to add them now; it'll be faster for you to iterate without them.

.github/workflows/on-push-to-main.yml
cprice404: i'm fine with trying what you have above. the stuff we had before was working so probably sticking closer to that will be less disruptive than changing to a new plugin

cprice404 (commented):
just a few comments

Cargo.toml
anitarua: So I think the utils file pulls in the [`CARGO_PKG_VERSION` env var](https://github.com/momentohq/client-sdk-rust/blob/main/src/utils.rs#L13) to use in the [`agent` header](https://github.com/momentohq/client-sdk-rust/blob/main/src/utils.rs#L104). And my understanding is that `CARGO_PKG_VERSION` gets pulled from this `.toml` file so I figured I'd have release please update the version number here

anitarua: though on a second look at the dart sdk, I think release please will already update version numbers in the main manifest file. Will remove this addition for now

cprice404: > So I think the utils file pulls in the [`CARGO_PKG_VERSION` env var](https://github.com/momentohq/client-sdk-rust/blob/main/src/utils.rs#L13) to use in the [`agent` header](https://github.com/momentohq/client-sdk-rust/blob/main/src/utils.rs#L104). And my understanding is that `CARGO_PKG_VERSION` gets pulled from this `.toml` file so I figured I'd have release please update the version number here

Oh I understand what you're saying now. Yes we do want it to update this so feel free to add it back in if you end up needing to.

cprice404 (approved):

PR #191, opened 2024-03-21
feat: tweaks to builders

cprice404:
Cleans up the builder / constructor interfaces for Configuration, CredentialProvider, and CacheClient.

Specifically:

* Introduce phased builders for Configuration and CacheClient to try to make sure we have flexibility if we want to add new optional arguments in the future, or make some of the current arguments optional.
* Make all of the builder constructors take no args, for consistency
* Get rid of the builder struct for CredentialProvider entirely; it was already private, but after making the `from_env_var` and `from_string` functions responsible for parsing the tokens and returning a MomentoResult, there was nothing left for the builder to do. This also resulted in allowing us to pull the endpoint override logic up and separate it from the token parsing logic, which cleans up the code a bit.
* Added 'base_endpoint' override to CredentialProvider. We can add back the ability to override specific endpoints in the near future.
* Added a placeholder for the token endpoint to CredentialProvider.


nand4011 (approved):
LGTM. I'm still trying to figure out if I prefer whether functions that take a field and return a struct with the new field start with `with_`. It only sticks out to me in the `base_endpoint` function of the credential provider. It might be my getter/setter naming instincts bugging me unnecessarily.

cprice404:
> LGTM. I'm still trying to figure out if I prefer whether functions that take a field and return a struct with the new field start with `with_`. It only sticks out to me in the `base_endpoint` function of the credential provider. It might be my getter/setter naming instincts bugging me unnecessarily.

Yeah it's a tricky one. In my experience it implies a copy constructor and I don't want to imply that in most of these since we may be mutating rather than creating a copy of the struct, but I don't know if my association of `with` to copy constructors is well-founded or not.

fwiw i am also taking some cues from the other builders i've seen in libs like the aws sdk and the hyper/tonic stuff, and they are not using a prefix like that, so... slightly tilted scales for me

example/src/bin/readme.rs
```diff
@@ -7,11 +7,13 @@ const CACHE_NAME: &str = "cache";
...
-        CredentialProvider::from_env_var("MOMENTO_API_KEY".to_string())?,
-        laptop::latest(),
-        Duration::from_secs(60),
-    )?;
+    let cache_client = CacheClient::builder()
+        .default_ttl(Duration::from_secs(60))
+        .configuration(laptop::latest())
+        .credential_provider(CredentialProvider::from_env_var(
+            "MOMENTO_API_KEY".to_string(),
+        )?)
```
kvcache, line 15: this from_env_var function should take an `impl Into<String>` so that users do not need to manually convert `&'static str` into String. That said, I imagine you should just use an `&str` instead because I don't expect you have use of the environment variable after the function returns right?

scripts/test-setup.rs
```diff
@@ -10,12 +10,12 @@ async fn main() {
...
-        configurations::laptop::latest(),
-        Duration::from_secs(5),
-    )
-    .expect("cache client cannot be created");
+    let cache_client = CacheClient::builder()
+        .default_ttl(Duration::from_secs(5))
+        .configuration(configurations::laptop::latest())
+        .credential_provider(credential_provider)
+        .build()
+        .expect("cache client cannot be created");
```
kvcache, line 18: Nit: https://doc.rust-lang.org/std/result/enum.Result.html#method.expect

> If you’re having trouble remembering how to phrase expect error messages remember to focus on the word “should” as in “env variable should be set by blah” or “the given binary should be available and executable by the current user”.
> For more detail on expect message styles and the reasoning behind our recommendation please refer to the section on [“Common Message Styles”](https://doc.rust-lang.org/std/error/index.html#common-message-styles) in the [std::error](https://doc.rust-lang.org/std/error/index.html) module docs.

src/config/configuration.rs
```diff
@@ -20,22 +20,25 @@ use crate::config::transport_strategy::TransportStrategy;
...
-///        .build();
-#[derive(Clone)]
+/// let config = Configuration::builder()
+///     .transport_strategy(
+///         TransportStrategy::builder()
+///             .grpc_configuration(
+///                 GrpcConfiguration::builder()
+///                     .deadline(Duration::from_millis(1000))
+///                     .build()    
+///             ).build()
```
kvcache, line 30: `grpc_configuration()` should take an `impl Into<GrpcConfiguration>` and `GrpcConfigurationBuilder` should have From impl'd at the appropriate stage.

```
impl From<GrpcConfigurationBuilder<ReadyToBuild>> for GrpcConfiguration {
    fn from(self) -> GrpcConfiguration {
        self.build()
    }
}
```

so that the usage becomes
```
.grpc_configuration(
    GrpcConfiguration::builder()
        .deadline(Duration::from_millis(1000))
).build()
```

And then the transport_strategy function should take an `impl Into<TransportStrategy>`, and then there should be impl From<TransportStrategyBuilder<ReadyToBuild>> for TransportStrategy [...] which results in:

```
/// let config = Configuration::builder()
///     .transport_strategy(
///         TransportStrategy::builder()
///             .grpc_configuration(
///                 GrpcConfiguration::builder().deadline(Duration::from_millis(1000)
///             )
///     )
///     .build();
```

And if you accept `impl Into<Configuration>` in your client constructor then you can simplify most use cases to

```
/// let config = Configuration::builder()
///     .transport_strategy(
///         TransportStrategy::builder()
///             .grpc_configuration(
///                 GrpcConfiguration::builder().deadline(Duration::from_millis(1000)
///             )
///     );
```

kvcache (commented):
I think this is a great improvement

example/src/bin/readme.rs
cprice404: I have a different ticket about taking a pass over everything and cleaning up the String arg stuff, so I probably won't handle that in this PR, but - what would your general guidance be on this? Always prefer `&str` unless we need to take ownership? and if we need to take ownership then always prefer `impl Into<String>`?

scripts/test-setup.rs
cprice404: :ack:

src/config/configuration.rs
cprice404: nice, will try this.

example/src/bin/readme.rs
kvcache: * In tests, always take `impl Into<String>` because that gives you the most flexibility and you do not care about memcpy cost
* In setup code _probably_ take `impl Into<String>` because that bothers users as little as possible with whatever string-like thing they want to use to identify their environment variable. They don't have to & anything.
* In steady-state code, _always_ take `&str` when your function chain only needs to have the string as long as the function chain exists, or when the clone is only occasional.
* In steady-state code, _probably_ take `impl Into<String>` when an owned `String` will be required to fulfill the function's contract (e.g., when your function requires filling out a protocol buffers object, which requires owned `String`s). The other choice is accepting an owned `String`, which certain microcontroller developers might prefer because there would be a couple fewer bytes of generated code. It's unlikely to be a consideration for momento's current target deployment scenarios.

scripts/test-setup.rs
cprice404: fixed in https://github.com/momentohq/client-sdk-rust/pull/191/commits/44fb3db09595c3ee6b41fc20aebfa954198cdaaf

src/config/configuration.rs
cprice404: fixed in https://github.com/momentohq/client-sdk-rust/pull/191/commits/44fb3db09595c3ee6b41fc20aebfa954198cdaaf

kvcache (approved):
looks nice to me

example/src/bin/readme.rs
cprice404: this is awesome, thanks so much. I will go back and apply this to all of the APIs when I pick up that String ticket, very soon :)

PR #193, opened 2024-03-21
chore: minor re-organization of integration tests and utils

cprice404:
This commit just attempts to re-organize the int test file hierarchy a bit to make it easier to navigate in the future. Of note:

* Make a subdir for 'cache' integration tests so we can easily separate topic and other int tests in the future
* Move the SimpleCacheClient int tests into an `old` subdir to make it obvious they are going away
* Move the `cache_test_state` stuff up into the existing `test-util` dir, where the doctest helpers are defined, to consolidate all the utility/helper code into the same spot.


nand4011 (approved):

pratik151192 (approved):

test-util/Cargo.toml
```diff
@@ -10,3 +10,4 @@ anyhow = "1.0.68"
 tokio = { version = "1.25.0", features = ["full"] }
 uuid = { version = "1.2.2", features = ["v4"] }
 scopeguard = "1.1.0"
+once_cell = "1.19.0"
```
malandis, line 13: Nice catch

test-util/src/test_utils.rs
```diff
@@ -0,0 +1,73 @@
...
+
+            // If any background tasks were spawned we give them some time to exit cleanly.
+            runtime.shutdown_timeout(Duration::from_secs(1));
+        }
+    });
+
+    runtime.block_on(func(cache_name, credential_provider))
+}
+
+pub fn create_doctest_client() -> (CacheClient, String) {
```
malandis, line 52: Rename to `create_doctest_cache_client`?

malandis (commented):
One suggestion. Looks good otherwise 👍

test-util/src/test_utils.rs
cprice404: fixed in https://github.com/momentohq/client-sdk-rust/pull/193/commits/ae6f8e330f70d2990eee68d32bf8adb2f1760c0e

pgautier404 (approved):

PR #195, opened 2024-03-22
chore: move new sorted set fetch response type to new requests dir

cprice404:
This is just to keep the new files separated from the old files. The sorted_set_fetch response definition was inside the old `responses` dir, which i'm expecting us to delete before too long. This just moves it over to the new directory structure. We can reorganize a bit further after we delete the old code if necessary.


pratik151192 (approved):

anitarua (dismissed):

malandis (approved):

nand4011 (dismissed):

cprice404:
rebased

nand4011 (approved):

malandis (approved):

PR #196, opened 2024-03-22
chore: update CONTRIBUTING re: tests

cprice404:
The goal of this commit is to update the CONTRIBUTING.md file to include more information about how the different kinds of tests should be organized and how to run them.

I also tweaked the directory structure of the int tests again. I'm flattening it back out because unfortunately the `cargo test` CLI doesn't seem to have a good way to only run selected tests if they are nested in subdirs, though it has good support for it when they are defined in the root of the `tests` dir.


pratik151192 (dismissed):

anitarua (dismissed):

CONTRIBUTING.md
```diff
@@ -60,3 +60,64 @@ Run this command to verify everything passes so that you can save yourself some
 ```bash
 MOMENTO_API_KEY=<api key> make
 ```
+
+### Tests
+
+There are three different kinds of tests in this repo:
+
+* Unit tests: these are defined vi the `#[test]` macro, in the same file as the code they are testing. For an example see `credential_provider.rs`.
```
malandis, line 68:
```suggestion
* Unit tests: these are defined via the `#[test]` macro, in the same file as the code they are testing. For an example see `credential_provider.rs`.
```

CONTRIBUTING.md
```diff
@@ -60,3 +60,64 @@ Run this command to verify everything passes so that you can save yourself some
...
+
+### More details on doctests
+
+Docstring are denoted via `///` comment blocks preceding a function or struct.
+
+Inside of these docstrings you can include code blocks, via `/// \`\`\``.
+
+The code inside of these code blocks will be compiled and executed as part of the doctest execution. This ensures that our docs always have working code; the compile will fail if the docstring code is not valid.j
+
+The resulting docs will be automatically published to https://docs.rs/momento/latest/momento/ . However you may hide certain lines from the published docs by prefixing them with a `#` (`/// #`).
```
malandis, line 110:
```suggestion
The resulting docs will be automatically published to https://docs.rs/momento/latest/momento/ . However you may hide certain lines from the published docs by prefixing them with a `#` (`/// #`). Use this liberally to remove extraneous details from the documentation while keeping the code blocks self-contained and syntactically valid.
```

malandis (commented):
Looks good. Just a couple suggestions and one based on our talk yesterday.

nand4011 (dismissed):

cprice404:
rebased and addressed Michael's feedback.

malandis (approved):

nand4011 (approved):

Issue #171, opened 2024-02-15
Ensure that error messages are descriptive


nand4011:
Error messages should be as descriptive as the error messages in the JavaScript SDK: https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/errors/errors.ts

cprice404:
I haven't looked over the code for the errors in the Rust SDK yet but I would take this description a step further and say that we should try to make sure that the errors are modeled as closely as possible to the ones in the other SDKs. So we should make sure the list of error types in the enum / class hierarchy matches, make sure the messages are consistent, make sure we have properties for the errorCode / innerError, etc.

PR #197, opened 2024-03-22
fix: standardize error messages, add details

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/171

Trying to make the returned MomentoErrors more consistent with our other SDKs. Changed it from an enum to a struct with accessible `message`, `error_code`, `inner_error`, and GRPC `details` fields.

Most of the changes are in `response/errors.rs`, but I imagine we want to move that file somewhere else, I just didn't know where yet -> moved to `requests/errors.rs` for now.

The rest of the changes are mostly updating the places where we construct these errors to be returned.


src/response/error.rs
```diff
@@ -4,108 +4,124 @@ use tonic::codegen::http;
...
+    /// Unknown error has occurred
+    UnknownError,
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("{details}")]
+pub struct MomentoGrpcErrorDetails {
+    pub code: tonic::Code,
+    pub details: Cow<'static, str>,
+    pub metadata: Option<tonic::metadata::MetadataMap>,
```
anitarua, line 46: adding metadata produces errors like:

```
---- tests::ttl_validation stdout ----
thread 'tests::ttl_validation' panicked at tests/integration_test.rs:83:14:
failed to create cache: LimitExceeded(SdkError { message: "Request rate, bandwidth, or object size exceeded the limits for this account.  To resolve this error, reduce your usage as appropriate or contact us at support@momentohq.com to request a limit increase", error_code: LimitExceededError, inner_exception: Some(TonicStatus(Status { code: ResourceExhausted, message: "Your account has a limit of 10 caches. Please contact support if your use case necessitates more caches.", metadata: MetadataMap { headers: {"content-type": "application/grpc", "server": "awselb/2.0", "date": "Fri, 22 Mar 2024 17:04:33 GMT", "content-length": "0", "access-control-expose-headers": "grpc-status,grpc-message,grpc-encoding,grpc-accept-encoding", "vary": "origin", "vary": "access-control-request-method", "vary": "access-control-request-headers", "access-control-allow-credentials": "true"} }, source: None })), details: Some(MomentoGrpcErrorDetails { code: ResourceExhausted, details: "resource exhausted", metadata: Some(MetadataMap { headers: {"content-type": "application/grpc", "server": "awselb/2.0", "date": "Fri, 22 Mar 2024 17:04:33 GMT", "content-length": "0", "access-control-expose-headers": "grpc-status,grpc-message,grpc-encoding,grpc-accept-encoding", "vary": "origin", "vary": "access-control-request-method", "vary": "access-control-request-headers", "access-control-allow-credentials": "true"} }) }) })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Can remove the metadata if it seems like too much

src/response/error.rs
```diff
@@ -4,108 +4,124 @@ use tonic::codegen::http;
...
+
+#[derive(Debug, thiserror::Error)]
+#[error("{message}")]
+pub struct SdkError {
+    pub message: Cow<'static, str>,
+    pub error_code: MomentoErrorCode,
+    #[source]
+    pub inner_exception: Option<ErrorSource>,
+    pub details: Option<MomentoGrpcErrorDetails>,
+}
```
anitarua, line 57: Instead of specifying the fields for each MomentoError enum case, I made a struct that each error should adhere to instead (with the exception of Miss and TypeError, weren't sure about those).

Not exactly certain about the types specified here but these fields are at least consistent with the other SDKs

src/credential_provider.rs
```diff
@@ -300,20 +305,25 @@ impl CredentialProviderBuilder {
...
+                error_code: MomentoErrorCode::InvalidArgumentError,
+                inner_exception: None,
+                details: None
+            }));
         Ok(CredentialProvider {
             auth_token,
-            cache_endpoint: CredentialProviderBuilder::https_endpoint(cache_endpoint),
-            control_endpoint: CredentialProviderBuilder::https_endpoint(control_endpoint),
+            cache_endpoint: CredentialProviderBuilder::https_endpoint(cache_endpoint?),
+            control_endpoint: CredentialProviderBuilder::https_endpoint(control_endpoint?),
```
anitarua, line 326: code hint suggested using `?` here but I'm not entirely sure yet this is correct

anitarua: TIL: https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html

cprice404:
> Addresses #171
> 
> Trying to make the returned MomentoErrors more consistent with our other SDKs. Most of the changes are in `response/errors.rs`, but I imagine we want to move that file somewhere else, I just didn't know where yet.

You can just move it into `requests/errors.rs` for now. We can re-organize after we delete the old stuff.

src/auth.rs
```diff
@@ -67,30 +67,42 @@ pub enum AuthError {
...
-                source: err.into(),
-            },
-            AuthError::ActionError(_) => MomentoError::ClientSdkError {
-                description: "login action failed".into(),
-                source: err.into(),
-            },
+            AuthError::LoginAborted => MomentoError::Cancelled(SdkError {
+                message: "aborted login".into(),
+                error_code: MomentoErrorCode::CancelledError,
+                inner_exception: Some(err.into()),
```
cprice404, line 73: supernit but in rust there really isn't such a thing as an exception, so let's call this field `inner_error` maybe?

src/response/error.rs
```diff
@@ -147,10 +163,12 @@ pub enum ErrorSource {
 
 impl From<String> for MomentoError {
```
cprice404, line 164: where does this get used? It seems sketchy to assume that anytime someone wants to convert a String to a MomentoError that it would always be a `BadRequest`? I'm not sure it even makes sense to me to implement `From<String>` for MomentoError, but if it did make sense I would expect it to map to like an UnknownError.

src/response/error.rs
```diff
@@ -163,50 +181,116 @@ impl From<tonic::Status> for MomentoError {
...
+                metadata: Some(status.metadata().clone())
+            }.into()
+        }),
+        tonic::Code::Unimplemented => MomentoError::BadRequest(SdkError {
+            message: "The request was invalid; please contact us at support@momentohq.com".into(),
+            error_code: MomentoErrorCode::BadRequestError,
+            inner_exception: Some(status.clone().into()),
+            details: MomentoGrpcErrorDetails {
+                code: status.code(),
+                details: "unimplemented".into(),
```
cprice404, line 200: it looks to me like the `status` object has a function called `details`, I think you will want to call that rather than making up our own.

There's also a `message` fn. We should add a corresponding `message` field to ours and expose that.

Also I don't think that converting from `status` to these `MomentoGrpcErrorDetails` should differ at all from one case to the next, so I think you can define `impl From<tonic::Status> for MomentoGrpcErrorDetails` and then all of these will just be `details: Some(status.into())`, just like what you are doing for the inner_exception.

src/response/error.rs
```diff
@@ -4,108 +4,124 @@ use tonic::codegen::http;
...
+    FailedPreconditionError,
+    /// Unknown error has occurred
+    UnknownError,
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("{details}")]
+pub struct MomentoGrpcErrorDetails {
+    pub code: tonic::Code,
+    pub details: Cow<'static, str>,
```
cprice404, line 45: I think this should just be a String.

src/response/error.rs
cprice404, line 46: It looks to me like the tonic `Status` struct always has a Metadata on it, so I think this should not be `Option`?

src/response/error.rs
```diff
@@ -4,108 +4,124 @@ use tonic::codegen::http;
...
+    pub metadata: Option<tonic::metadata::MetadataMap>,
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("{message}")]
+pub struct SdkError {
+    pub message: Cow<'static, str>,
```
cprice404, line 52: I think this should also be `String`.

src/response/error.rs
cprice404: I think that is okay! We're giving people lots of ways to access more precise bits of info from the error if they want to.

src/response/error.rs
cprice404: I like this but then I think the way you're defining the stuff in the enum below, you are basically defining each Error type as a 1-item-tuple. (I think. I'm still learning some of this rust syntax myself.)

For such tuples I've seen some weird syntax where you'd have to write something like `error.0.description` to access the description field, and we don't want that. But in your case since it's an enum they might have to do a `match` first, and maybe they would be destructuring the inner error in the `match`, so maybe it's okay. is there some code somewhere in this PR that illustrates what it looks like to handle one of these errors and access one of the fields inside?

I'm also wondering about whether it's possible for someone to have a reference to a MomentoError and then access the inner fields without having to `match` to a specific value from the enum first. I wonder if we can do like `impl MomentoError { ... }` and provide some common accessors that can be used without the match. I don't think it works that way but it'd be really desirable from an end user perspective so we might want to play around and look for some way to pull that off. Like, maybe `MomentoError` becomes a struct with the fields that you currently put into `SdkError`, and then we add a field to it called `error_type` or something, and the value of that is the enum?

We can chat about this and trade ideas if you like.

cc: @kvcache you will probably have opinions about this part.

cprice404 (commented):
nice work, I think this is on the right track but left a few questions/comments.

src/response/error.rs
anitarua: not used anywhere apparently, will remove

src/response/error.rs
anitarua: Been testing locally to get an idea of the current UX. Was able to use [`if let`](https://doc.rust-lang.org/book/ch06-03-if-let.html) to match on the specific MomentoError enum type and then get at the SdkError struct by creating another variable:

```
let result = cache_client.create_cache(CACHE_NAME.to_string()).await;
match result {
    Ok(_) => println!("Successfully created cache '{}'", CACHE_NAME),
    Err(e) => if let MomentoError::AlreadyExists(already_exists) = e {
        println!("Cache already exists: {} | Metadata: {:?}", already_exists.message, already_exists.details.unwrap().metadata);
    } else if let MomentoError::NotFound(not_found) = e {
        println!("Cache '{}' not found | Metadata: {:?}", not_found.message, not_found.details.unwrap().metadata);
    } else {
        println!("Error: {}", e);
    }
}
```

src/requests/errors.rs
```diff
@@ -0,0 +1,270 @@
...
+
+#[derive(Debug, thiserror::Error)]
+#[error("{message}")]
+pub struct MomentoError {
+    pub message: String,
+    pub error_code: MomentoErrorCode,
+    #[source]
+    pub inner_error: Option<ErrorSource>,
+    pub details: Option<MomentoGrpcErrorDetails>,
+}
```
anitarua, line 73: Switched MomentoError from an enum to a struct with publicly accessible methods. Example of the UX:

```
let result = cache_client.create_cache(CACHE_NAME.to_string()).await;
    match result {
    Ok(_) => println!("Successfully created cache '{}'", CACHE_NAME),
    Err(e) => println!("Error code: {:?} | Message: {}", e.error_code, e.message)
}
```

cprice404: I think I like that - can you still do an `if let` on the `Err` branch like you did in your code in your earlier comment above?

Also I just noticed the `AlreadyExists` stuff above, which we moved to be a non-error case in the CreateCache response in the other SDKs, probably worth getting that in while we are messing with this.

cprice404 (commented):
left a couple of quick questions/comments; lmk what you think about those and then I will pull this down and play around with the ux of it.

src/requests/errors.rs
anitarua: Didn't remove the AlreadyExists error completely since that can still be an error code received from GRPC, but the CreateCache response is now an enum and allows the possibility of checking if the Ok response was Created or AlreadyExists:

```
let result = cache_client.create_cache(CACHE_NAME.to_string()).await;
if let Err(e) = result {
    println!("Error: {}", e);
} else if let Ok(ok) = result {
    match ok {
        CreateCache::Created => println!("Successfully created cache '{}'", CACHE_NAME),
        CreateCache::AlreadyExists => println!("Cache '{}' already exists", CACHE_NAME),
    }
}
```

cprice404: > Makes sense, will remove `AlreadyExists` from the errors list.
> 
> Using `if let` could look something like this:
> 
> ```
> let result = cache_client.create_cache(CACHE_NAME.to_string()).await;
> if let Err(e) = result {
>     println!("Error: {}", e);
> } else {
>     println!("Successfully created cache '{}'", CACHE_NAME);
> }
> ```

Cool, thanks. I'm also interested in seeing if there's a way to do an if-let on the specific error type (maybe just the error code now), like you had done in your first example.

i.e., in your first example, I really liked how you could write conditions for the specific error types if you cared about them, but it seemed maybe hard to interact with errors generically. Now I think you've solved the "generically" case, and I want to make sure it's still also possible to drill in on the specific types if users want to.

anitarua: Yeah seems possible to use `if let` on the error codes instead:

```
let result = cache_client.create_cache(CACHE_NAME.to_string()).await;
if let Err(e) = result {
    if let MomentoErrorCode::InternalServerError = e.error_code {
        println!("InternalServerError: {}", e);
        return Ok(());
    } else {
        println!("Error: {}", e);
    }
} else if let Ok(ok) = result {
    match ok {
        CreateCache::Created => println!("Successfully created cache '{}'", CACHE_NAME),
        CreateCache::AlreadyExists => println!("Cache '{}' already exists", CACHE_NAME),
    }
}
```

src/requests/errors.rs
```diff
@@ -0,0 +1,270 @@
...
+use crate::auth::AuthError;
+
+#[derive(Debug, Clone)]
+pub enum MomentoErrorCode {
+    /// Invalid argument passed to Momento client
+    InvalidArgumentError,
+    /// Service returned an unknown response
+    UnknownServiceError,
+    /// Resource with specified name already exists
+    AlreadyExistsError,
```
cprice404, line 14: we want to take this out of this enum for now I think? check the other sdks

src/requests/errors.rs
cprice404: i pulled this code down and played with it, and you can match on e.error_code too, so overall I think I am liking this pattern a lot better. thanks!

src/response/cache/sorted_set_fetch.rs
```diff
@@ -0,0 +1,138 @@
...
+use momento_protos::cache_client::sorted_set_fetch_response::found::Elements;
+use momento_protos::cache_client::sorted_set_fetch_response::SortedSet;
+use momento_protos::cache_client::SortedSetFetchResponse;
+
+use crate::{
+    response::{MomentoErrorCode, SdkError},
+    MomentoError, MomentoResult,
+};
+
+// TODO this needs to be moved to the requests directory
```
cprice404, line 12: i think i moved this in my PR that got merged, and it looks like maybe during rebase you re-introduced it?

cprice404 (commented):
i'm pretty happy with this now! left a few more nits about whether that AlreadyExists should still be in the enum, and whether you might have accidentally re-introduced that sorted set file in the `responses` dir, but I'm 🚢 otherwise

src/response/cache/sorted_set_fetch.rs
anitarua: oh whoops, will fix that

src/requests/errors.rs
cprice404: well. TIL. okay 👍

cprice404 (approved):
clap

cprice404:
@malandis do you want to look at this before it goes in?

PR #201, opened 2024-03-25
feat: use impl Into<String> on most public APIs

cprice404:
This commit attempts to take a pass through all of the public APIs and find places where we were unnecessarily being too rigid about what flavor of String we accept. We use `impl Into<String>` as the argument type in most places now, which means that callers can pass us either `&str` or `String`.


anitarua (dismissed):

malandis (dismissed):

cprice404:
rebased

anitarua (approved):

PR #202, opened 2024-03-26
docs: document gotcha about multiple rust toolchain installation

malandis:
We add a block to the CONTRIBUTING section warning about a common
pitfall when there are multiple rust installations.


bruuuuuuuce (approved):
:shipit:

anitarua (approved):

PR #207, opened 2024-03-27
feat: add `IntoSortedSetElements` trait and implementations for Vec and HashMap

malandis:
This adds a `IntoSortedSetElements` trait to serve as an interface
between various collection types and the
`SortedSetPutElementsRequest`.

We provide implementations for `Vec` and `HashMap` and refactor the
code to consume the trait.


src/requests/cache/sorted_set/sorted_set_put_elements.rs
```diff
@@ -1,17 +1,104 @@
+use std::collections::HashMap;
+
 use momento_protos::cache_client::{SortedSetElement, SortedSetPutRequest};
 
 use crate::requests::cache::MomentoRequest;
 use crate::simple_cache_client::prep_request_with_timeout;
 use crate::{CacheClient, CollectionTtl, IntoBytes, MomentoResult};
 
+/// This trait defines an interface for converting a type into a vector of `SortedSetElement`s.
+pub trait IntoSortedSetElements: Send {
```
malandis, line 10: I put this in this file and at the top. From what I've read the convention is to put traits before the things that consume them in the file. Let me know if this should be in its own file.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
```diff
@@ -1,17 +1,104 @@
...
+/// #        value: "value1".as_bytes().to_vec(),
+/// #        score: 1.0,
+/// #    },
+/// #    momento_protos::cache_client::SortedSetElement {
+/// #        value: "value2".as_bytes().to_vec(),
+/// #        score: 2.0,
+/// #    },
+/// # ]);
+/// ```
+pub fn map_and_collect_sorted_set_elements<I, V>(iter: I) -> Vec<SortedSetElement>
```
malandis, line 70: This will be helpful to anyone who implements for their own collection type, hence I made public.

I wanted this to be another default function on the trait, but I could not get the `IntoIterator<(IntoBytes, f64)>` type to work as I wanted.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
```diff
@@ -61,24 +148,18 @@ impl<S: IntoBytes, E: IntoBytes> SortedSetPutElementsRequest<S, E> {
 
     /// Set the time-to-live for the collection.
     pub fn with_ttl(mut self, collection_ttl: CollectionTtl) -> Self {
+        // TODO don't we want this to return a copy as opposed to mutate self?
```
malandis, line 151: I noticed this method was called `with_ttl`, which makes it sound like a builder (hence copy constructor semantics), but it is instead mutating `self`. Was this intended or is it desired?

anitarua: Definitely seems like it'd be more consistent with the other SDKs to make this return a copy instead

anitarua (dismissed):

src/requests/cache/sorted_set/sorted_set_put_elements.rs
cprice404: if this is the only file where this will be consumed then this is good. if you think it might be used in any other spots then we should probably pull it out?

src/requests/cache/sorted_set/sorted_set_put_elements.rs
cprice404: let's keep this private for now. if you can figure out a way to do an Into then I'm okay with it being public (and I will look at the code a bit in a second) but I don't want to add utility functions to our public API surface area out of the gate.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
malandis: I only expect it to be used externally (tests and by developers).

src/requests/cache/sorted_set/sorted_set_put_elements.rs
malandis: changed in [f5cd3b2](https://github.com/momentohq/client-sdk-rust/pull/207/commits/f5cd3b25d6fd398a6dfcb18be37b8228d17d1e4c)

src/requests/cache/sorted_set/sorted_set_put_elements.rs
cprice404: there isn't a precedent for this in any of the SDKs except for the golang one, because we don't expose request objects in any other sdks.

I don't really imagine that users are going to be keeping around a request object that they want to use as a template and continue to create clones of it via the copy constructor, and if they do want to do that, rust provides `.clone`. (side note, we should make sure all of our request objects derive clone).

rust users in general are more concerned with not wasting memory or cpu cycles by copying things that don't need to be copied, so my inclination is that we should err on the side of not wasting the memory to make the copies.

I brought up the `with_` question in sprint planning because I have had this same conversation with @nand4011 a few times now. I think it comes down to:

1. Will rust users interpret the `with_` prefix to imply a copy constructor? Or am I (slash are we) projecting that based on our work in our other SDKs?
2. If we believe that people will interpret this as a copy constructor, then should we remove the prefix? (https://github.com/momentohq/dev-eco-issue-tracker/issues/738 ) That will come at the cost of our properties/options no longer being grouped together in the autocomplete hints in the IDEs. I tried that out and it feels weird. That is why I ended up coming back around to the opinion that we should keep the prefix.

I don't think these should be copy constructors, especially for the request objects. I'm open to removing the prefix or coming up with a different one.

Thoughts?

src/requests/cache/sorted_set/sorted_set_put_elements.rs
```diff
@@ -1,17 +1,104 @@
...
+/// matches the order of pairs in the input iterator.
+///
+/// # Examples
+///
+/// Basic usage with a vector of tuples:
+///
+/// ```
+/// use momento::requests::cache::sorted_set::sorted_set_put_elements::map_and_collect_sorted_set_elements;
+/// let pairs = vec![("value1", 1.0), ("value2", 2.0)];
+/// let sorted_set_elements = map_and_collect_sorted_set_elements(pairs.into_iter());
```
cprice404, line 35: Can they not just call `pairs.into()` here?

src/requests/cache/sorted_set/sorted_set_put_elements.rs
```diff
@@ -1,17 +1,104 @@
...
+/// #    momento_protos::cache_client::SortedSetElement {
+/// #        value: "value1".as_bytes().to_vec(),
+/// #        score: 1.0,
+/// #    },
+/// #    momento_protos::cache_client::SortedSetElement {
+/// #        value: "value2".as_bytes().to_vec(),
+/// #        score: 2.0,
+/// #    },
+/// # ]);
+/// ```
```
cprice404, line 69: these are great docs and I want to make sure we keep / expose them. I'm just hoping we can do it based on some flavor of `into` rather than the utility function. I am obviously missing something as to why that won't work, so I'll pull this branch down after I finish reviewing the PR and get myself up to speed :)

cprice404 (commented):
this looks awesome overall, just a few bits i don't understand yet. will pull the branch down and poke at it.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
malandis: Yeah but this is documentation of the utility method not the trait. I put this here in detail for anyone else who may wish to implement the trait for their own custom container.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
```diff
@@ -41,15 +128,15 @@ use crate::{CacheClient, CollectionTtl, IntoBytes, MomentoResult};
 /// # Ok(())
 /// # })
 /// # }
-pub struct SortedSetPutElementsRequest<S: IntoBytes, E: IntoBytes> {
+pub struct SortedSetPutElementsRequest<S: IntoBytes, E: IntoSortedSetElements> {
```
cprice404, line 131: you should also update the signature in `cache_client.sorted_set_put_elements`, that's actually where we get the most value from this

src/requests/cache/sorted_set/sorted_set_put_elements.rs
```diff
@@ -1,17 +1,104 @@
...
+    V: IntoBytes,
+{
+    iter.map(|(value, score)| SortedSetElement {
+        value: value.into_bytes(),
+        score,
+    })
+    .collect()
+}
+
+impl<V: IntoBytes> IntoSortedSetElements for Vec<(V, f64)> {
```
cprice404, line 80: i think you need to add one of these for `Vec<SortedSetElement>` too, otherwise callers can't pass that if they happen to already have one. Might be a rare case but I would expect this to work:

```rust
let sse = my_vec.into_sorted_set_elements();
client.sorted_set_put_elements(..., sse);
```

cprice404: oh hrm. i'm just now realizing that SortedSetElement is part of the protos and feeling a little iffy about whether it should be part of the public API or not. 🤔

src/requests/cache/sorted_set/sorted_set_put_elements.rs
cprice404, line 10: hmm, TIL; I was wondering why we are doing stuff this way for, e.g., IntoBytes; like, why aren't we just doing `impl From<Vec<(V, f64)>> for Vec<SortedSetElement>`?

Then I tried to impl that and it looks like the reason is that rust won't let you define an impl for a trait if both the trait and the struct you are trying to impl for are not defined in the current crate. That's a little bit of a bummer. Would be nice to just be able to do `.into()`, but, not a huge deal.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
malandis: Checking out some popular crates that use the builder pattern:
- to mutate a field, call the method the same as the field name
- to set or unset a boolean, can get creative with verbs

[Tonic builder](https://docs.rs/tonic-build/latest/tonic_build/)
[CSV writer builder](https://docs.rs/csv/latest/csv/struct.WriterBuilder.html)

These do have the word "builder" in the name. So maybe that helps clarify expectations too. At a minimum I vote for removing `with_`

src/requests/cache/sorted_set/sorted_set_put_elements.rs
cprice404: based on my comment above about my attempt to do `From`, I think the compiler won't let people impl this trait for other collection types. which seems like a bummer... I hope I'm missing something.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
cprice404: we can re-export it from the momento crate if it needs to surface as public API. I'm not sure that it does yet but gtk.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
malandis: Agreed. I tried a few alternatives thinking about how much boilerplate the interface introduces. Eg what we really want is for the trait to apply if your type has an iterator over tuples `(IntoBytes, f64)`. Then our trait would have a default method to convert from such an iterator to `Vec<SortedSetPutElement>`.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
malandis: I agree with our stance in other SDKs to draw a firm line separating the gRPC/internal types and the public API.

malandis: Hmm that would mean changing the trait to not use `SortedSetElement` either.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
cprice404: but where are you getting the assumption that `with_` implies copy constructor? that is the assumption I'm trying to challenge in my own brain.

Also - are you unconcerned about the fact that our properties won't be grouped together in autocomplete?

src/requests/cache/sorted_set/sorted_set_put_elements.rs
malandis: Thanks for the catch. Fixed in [ee60e18](https://github.com/momentohq/client-sdk-rust/pull/207/commits/ee60e18a7bf3b7b0fa2ba2a8bd4cf510a9dcda16)

src/requests/cache/sorted_set/sorted_set_put_elements.rs
malandis: > but where are you getting the assumption that with_ implies copy constructor? that is the assumption I'm trying to challenge in my own brain.

By our consistent usage of it like this in every other SDK. Just poking around at some other crates it seems the builder is a common pattern and they default to not use prefixes.

> Also - are you unconcerned about the fact that our properties won't be grouped together in autocomplete?
They are in vscode?

image

src/requests/cache/sorted_set/sorted_set_put_elements.rs
cprice404: yeah, or at least re-export it

malandis: Should we reconsider, the `lib.rs` re-exports various sorted set types defined in the protos.

cprice404: the lib.rs and mod.rs files are going to be gutted very soon. i will have to look and see what they are re-exporting and why, and we can decide from there. probably a good idea for us to have a huddle on friday morning to try to align on a few of these topics.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
malandis: I also confirmed not possible due to the reasons you mentioned below.

tests/cache_sorted_set.rs
```diff
@@ -59,37 +65,76 @@ async fn sorted_set_put_element_nonexistent_cache() {
...
 #[tokio::test]
 async fn sorted_set_put_elements_happy_path() {
+    async fn test_put_elements_happy_path(
+        client: Arc<CacheClient>,
+        cache_name: String,
+        to_put: impl IntoSortedSetElements<String> + Clone,
+    ) {
+        let sorted_set_name = "sorted-set-".to_string() + &Uuid::new_v4().to_string();
+        let result = client
+            .sorted_set_fetch_by_score(cache_name.clone(), sorted_set_name.clone(), Ascending)
```
anitarua, line 81: is it possible to avoid cloning cache_name and sorted_set_name in the tests?

malandis: The way we've written the cache client, the methods take ownership of all the arguments. So when we want to re-use the cache name (sorted set name), we have to clone first.

Looking at the client, it looks like this is the case for `cache_name` (`impl Into<String>`) and the collection names (`impl IntoBytes`).

@cprice404 have we discussed elsewhere about what convention we want for variables that will get passed around a lot like the cache name or collection name? Seems we would want a reference for those to not waste effort copying many times.

src/requests/cache/sorted_set/sorted_set_put_elements.rs
```diff
@@ -127,33 +140,37 @@ impl<V: IntoBytes> IntoSortedSetElements for HashMap<V, f64> {
...
 /// # })
 /// # }
-pub struct SortedSetPutElementsRequest<S: IntoBytes, E: IntoSortedSetElements> {
+pub struct SortedSetPutElementsRequest<S: IntoBytes, V: IntoBytes, E: IntoSortedSetElements<V>> {
     cache_name: String,
     sorted_set_name: S,
     elements: E,
     collection_ttl: Option<CollectionTtl>,
+    // V is only used for the `IntoSortedSetElement`'s generic type parameter.
+    _marker: PhantomData<V>,
```
cprice404, line 149: 👻

tests/cache_sorted_set.rs
cprice404: +1

cprice404 (commented):
the PhantomData thing is pretty wacky but I pulled the branch down and I don't see a better way 🤷 

The rest lgtm, except +1 on Anita's question about whether the clones are required in the tests.

tests/cache_sorted_set.rs
cprice404: can you make your test's cache_name argument be `impl Into<String>`?

Can you make your sorted_set_name variable be `&str`?

malandis: I was able to remove the clones on the cache name by passing references in [1b1d25c](https://github.com/momentohq/client-sdk-rust/pull/207/commits/1b1d25ca55f28967395bb1847c9df86fdb9f81c5).

For the sorted set name, I would have to create two variables, one with the `String` we build, then another that is a string reference. The alternative is in [722f53a](https://github.com/momentohq/client-sdk-rust/pull/207/commits/722f53a71fbb510d78cf042f23101eb021ffd58d) which calls `as_str` on each invocation.

anitarua (approved):

tests/cache_sorted_set.rs
```diff
@@ -73,23 +73,23 @@ fn compare_by_first_entry(a: &(String, f64), b: &(String, f64)) -> std::cmp::Ord
...
         client: Arc<CacheClient>,
-        cache_name: String,
+        cache_name: &String,
         to_put: impl IntoSortedSetElements<String> + Clone,
     ) {
-        let sorted_set_name = "sorted-set-".to_string() + &Uuid::new_v4().to_string();
+        let sorted_set_name = format!("sorted-set-{}", Uuid::new_v4());
         let result = client
-            .sorted_set_fetch_by_score(cache_name.clone(), sorted_set_name.clone(), Ascending)
+            .sorted_set_fetch_by_score(cache_name, sorted_set_name.as_str(), Ascending)
```
cprice404, line 81: this is a supernit but I would recommend `&sorted_set_name` here if it works. and if it doesn't work then I'd doublecheck our API sig and see if we think it *should* work :)

cprice404 (commented):
i left one more supernit but you should feel free to merge this whenever you and @anitarua are aligned, don't block on me.

tests/cache_sorted_set.rs
cprice404: Sorry, I missed that reply. I don't quite understand though, what error do you get if you just do `&sorted_set_name` on the function arg?

Don't block on this though. merge if you're confident.

anitarua:
I've synced the branch with main and made the tests more consistent with its usage of references instead of cloning and `?` and such.

cprice404 (approved):
:ty:

PR #208, opened 2024-03-27
chore: add happy path test for set API

pratik151192:
Just wanted to add a super simple test for `set` to see the behavior and get a feel of the package.

### DIFF
--- a/tests/scalar.rs
+++ b/tests/scalar.rs
@@ -1,0 +1,18 @@
+use momento::requests::cache::basic::set::Set;
+use momento_test_util::CACHE_TEST_STATE;
+use uuid::Uuid;
+
+#[tokio::test]
+async fn set_happy_path() {
+    let client = CACHE_TEST_STATE.client.clone();
+    let cache_name = CACHE_TEST_STATE.cache_name.clone();
+    let key = &Uuid::new_v4().to_string();
+    let value = &Uuid::new_v4().to_string();
+
+    let result = client
+        .set(cache_name.clone(), key.clone(), value.clone())
+        .await
+        .expect("Failed to execute set operation");
+
+    assert_eq!(result, Set {});
+}


anitarua (approved):

tests/scalar.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[tokio::test]
+async fn set_happy_path() {
+    let client = CACHE_TEST_STATE.client.clone();
+    let cache_name = CACHE_TEST_STATE.cache_name.clone();
+    let key = &Uuid::new_v4().to_string();
+    let value = &Uuid::new_v4().to_string();
+
+    let result = client
+        .set(cache_name.clone(), key.clone(), value.clone())
```
malandis, line 13: Is it necessary to clone the cache_name, key, and values here?

cprice404: it shouldn't be; if it is then our API needs improvement :P

tests/scalar.rs
```diff
@@ -0,0 +1,18 @@
+use momento::requests::cache::basic::set::Set;
+use momento_test_util::CACHE_TEST_STATE;
+use uuid::Uuid;
+
+#[tokio::test]
+async fn set_happy_path() {
+    let client = CACHE_TEST_STATE.client.clone();
+    let cache_name = CACHE_TEST_STATE.cache_name.clone();
+    let key = &Uuid::new_v4().to_string();
+    let value = &Uuid::new_v4().to_string();
```
cprice404, line 10: what is the type returned by `new_v4`?

tests/scalar.rs
```diff
@@ -0,0 +1,18 @@
+use momento::requests::cache::basic::set::Set;
+use momento_test_util::CACHE_TEST_STATE;
+use uuid::Uuid;
+
+#[tokio::test]
+async fn set_happy_path() {
+    let client = CACHE_TEST_STATE.client.clone();
+    let cache_name = CACHE_TEST_STATE.cache_name.clone();
```
cprice404, line 8: do you need to clone these? if so it could be a code smell in our API, we want to try to minimize the places that people need to clone things. We should be cloning things ourselves when it's necessary so that callers don't have to.

anitarua: Cloning was the pattern that was in `cache_sorted_set` but TIL borrowing would work fine too:
```
let client = &CACHE_TEST_STATE.client;
let cache_name = &CACHE_TEST_STATE.cache_name;
```

tests/scalar.rs
pratik151192: `cache_name` wasn't needed; but see an error if i don't clone other two:

```
 ---                  ^^^^^ the trait `From<&String>` is not implemented for `Vec<u8>`, which is required by `&String: IntoBytes`
```

tests/scalar.rs
pratik151192: `uuid`

tests/scalar.rs
nand4011: The examples and tests for the existing APIs all clone the arguments, so I followed that pattern. I don't have a good instinct for what would be idiomatic here. Do rust functions of this sort generally expect to take ownership of the arguments they are given?

cprice404: they do take onwership, but if you pass a `&str` then the ownership is just of a reference so you can still use it. i merged a PR this morning that cleaned up tests a bit to store some vars of type `&str` to make the tests a little simpler (i hope).

malandis:
This has been addressed in ongoing development.

PR #209, opened 2024-03-27
feat: add FlushCache and tests

anitarua:
adds FlushCache and a couple integration tests

### DIFF
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -13,6 +13,7 @@ use crate::requests::cache::basic::set::{Set, SetRequest};
 use crate::requests::cache::create_cache::{CreateCache, CreateCacheRequest};
 use crate::requests::cache::delete_cache::{DeleteCache, DeleteCacheRequest};
 use crate::requests::cache::list_caches::{ListCaches, ListCachesRequest};
+use crate::requests::cache::flush_cache::{FlushCache, FlushCacheRequest};
 use crate::requests::cache::set::set_add_elements::{SetAddElements, SetAddElementsRequest};
 use crate::requests::cache::sorted_set::sorted_set_fetch_by_rank::{
     SortOrder, SortedSetFetchByRankRequest,
@@ -136,7 +137,7 @@ impl CacheClient {
     }
 
     /// Lists all caches in your account.
-    ///
+    /// 
     /// # Examples
     /// Assumes that a CacheClient named `cache_client` has been created and is available.
     /// ```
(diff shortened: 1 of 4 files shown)


src/cache_client.rs
```diff
@@ -159,6 +160,52 @@ impl CacheClient {
...
+    /// # tokio_test::block_on(async {
+    /// use momento::requests::cache::flush_cache::FlushCache;
+    /// use momento::requests::cache::flush_cache::FlushCacheRequest;
+    /// # let (cache_client, cache_name) = create_doctest_cache_client();
+    ///
+    /// let flush_cache_request = FlushCacheRequest::new(cache_name.to_string());
+    ///
+    /// match cache_client.send_request(flush_cache_request).await {
+    ///     Ok(_) => println!("Flushed cache: {}", cache_name),
+    ///     Err(e) => eprintln!("Error flushing cache: {}", e),
```
malandis, line 199: I think it would help to see a match here on the error for `NOT_FOUND` vs everything else

src/requests/cache/flush_cache.rs
```diff
@@ -0,0 +1,63 @@
...
+/// use momento::requests::cache::flush_cache::FlushCacheRequest;
+/// # let (cache_client, cache_name) = create_doctest_cache_client();
+///
+/// match cache_client.flush_cache(cache_name.to_string()).await {
+///     Ok(_) => println!("Flushed cache: {}", cache_name),
+///     Err(e) => eprintln!("Error flushing cache: {}", e),
+/// }
+/// # Ok(())
+/// # })
+/// # }
```
malandis, line 29: We should look into rust doc features if there is hyperlinking or inlining in place so we don't have to repeat the code examples in multiple places (eg here and the cache client)

tests/cache_control.rs
```diff
@@ -1,4 +1,11 @@
 use momento::requests::cache::create_cache::CreateCache;
+use momento::requests::cache::{
+    basic::{
+        get::{Get, GetValue},
+        set::Set,
+    },
```
malandis, line 6: Did you mean to include these in this PR?

malandis (commented):
Looks good just a couple questions/comments for us going forward

tests/cache_control.rs
anitarua: yup, called Get and Set in the flushCache test below

src/requests/cache/flush_cache.rs
anitarua: I think there's a way to [link things by name](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html). I imagine we can probably link from the CacheClient to the individual request objects and provide different code examples that are relevant to each (e.g. the common use case call in the cache client and the `send_request` example in the request objects).

Will update here for FlushCache and do a sweep of that when working on https://github.com/momentohq/dev-eco-issue-tracker/issues/724 too

malandis: Thanks. I look forward to seeing it!

tests/cache_control.rs
malandis: 👍

src/requests/cache/flush_cache.rs
cprice404: yeah that sounds awesome, ty

tests/cache_control.rs
```diff
@@ -32,3 +39,56 @@ async fn lists_existing_test_cache() {
...
+#[tokio::test]
+async fn flush_existing_cache_returns_success() {
+    let client = CACHE_TEST_STATE.client.clone();
+    let cache_name = CACHE_TEST_STATE.cache_name.clone();
+
+    // Insert some elements
+    let set_result1 = client
+        .set(cache_name.to_string(), "key1", "value1")
+        .await
+        .unwrap();
```
cprice404, line 61: we generally want to try to avoid calling `unwrap`. It's better to give the test fns a return type of `MomentoResult<()>`, and then you can use `await?`. The test will need to return `Ok(())` at the end.

Our linter usually fails builds if we use `unwrap` but I guess it doesn't apply that rule to tests. It's not the end of the world to use them here but better to get out of the habit of using them I think.

tests/cache_control.rs
```diff
@@ -1,34 +1,95 @@
...
+};
+use momento::MomentoResult;
 use uuid::Uuid;
 
 use momento_test_util::CACHE_TEST_STATE;
 
 #[tokio::test]
-async fn delete_nonexistent_cache_returns_not_found() {
-    let client = CACHE_TEST_STATE.client.clone();
+async fn delete_nonexistent_cache_returns_not_found() -> MomentoResult<()> {
```
anitarua, line 15: tests now return an empty MomentoResult if successful

tests/cache_control.rs
```diff
@@ -1,34 +1,95 @@
...
+    Ok(())
+}
+
+#[tokio::test]
+async fn flush_existing_cache_returns_success() -> MomentoResult<()> {
+    let client = &CACHE_TEST_STATE.client;
+    let cache_name = &CACHE_TEST_STATE.cache_name;
+
+    // Insert some elements
+    let set_result1 = client.set(cache_name.to_string(), "key1", "value1").await?;
```
cprice404, line 63: I really want to figure out how to prevent users from needing to do this `to_string()` call here. This isn't an issue with your PR most likely, probably something with how we are defining the signature of the function. What error do you get if you don't do the `to_string()`?

cprice404 (commented):
looks good just one more question about consuming our API and the cache_name arg

tests/cache_control.rs
anitarua: it's fine without `to_string()`, might've been leftover from when `cache_name` was cloned instead of borrowed or something. will remove those `to_string()` calls

cprice404 (approved):
flush

PR #213, opened 2024-03-28
chore: improve docs, replace asserts with pattern matching, reduce duplication of code examples

anitarua:
Addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/724 and general docs improvements.

Code examples in the `cache_client.rs` file now show examples of `cache_client.*` calls but link to the specific request objects that show examples of `send_request` calls.

With the exception of the `cache_client::get` example, I removed the assert statements in favor of pattern matching examples. A couple of the pattern matching statements (`deleteCache` and `set`) show how to handle specific errors like `NotFoundError`. 

Not specifically attached to the style of the code examples, so if it's more idiomatic to show a simple API call with an `await?` at the end rather than using a `match` statements, I can change that.

### DIFF
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -208,7 +208,8 @@ impl CacheClient {
     /// # })
     /// # }
     /// ```
-    /// You can also use the [send_request](CacheClient::send_request) method to get an item using a [SetRequest].
+    /// You can also use the [send_request](CacheClient::send_request) method to get an item using a [SetRequest]
+    /// which will allow you to set [optional arguments](SetRequest#optional-arguments) as well.
     pub async fn set(
         &self,
         cache_name: impl Into<String>,
@@ -295,7 +296,8 @@ impl CacheClient {
     /// # })
     /// # }
     /// ```
-    /// You can also use the [send_request](CacheClient::send_request) method to get an item using a [SetAddElementsRequest].
+    /// You can also use the [send_request](CacheClient::send_request) method to get an item using a [SetAddElementsRequest] 
+    /// which will allow you to set [optional arguments](SetAddElementsRequest#optional-arguments) as well.
     pub async fn set_add_elements<E: IntoBytes>(
         &self,
         cache_name: impl Into<String>,
@@ -347,7 +349,8 @@ impl CacheClient {
     /// # })
     /// # }
     /// ```
-    /// You can also use the [send_request](CacheClient::send_request) method to get an item using a [SortedSetPutElementRequest].
+    /// You can also use the [send_request](CacheClient::send_request) method to get an item using a [SortedSetPutElementRequest] 
+    /// which will allow you to set [optional arguments](SortedSetPutElementRequest#optional-arguments) as well.
     pub async fn sorted_set_put_element(
         &self,
         cache_name: impl Into<String>,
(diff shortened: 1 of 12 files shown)


malandis:
I agree the docs should show how to use the pattern matching as a developer using the library would. This is a good change. With removing the assertions though, we lose the automated tests, which is less desirable.

If we plan for comprehensive integration tests, I am less concerned about the latter, as we can port the tests there.

src/cache_client.rs
```diff
@@ -62,30 +62,15 @@ impl CacheClient {
...
-    ///
-    /// let create_cache_request = CreateCacheRequest::new(cache_name);
-    ///
-    /// let create_cache_response = cache_client.send_request(create_cache_request).await?;
-    ///
-    /// assert_eq!(create_cache_response, CreateCache::AlreadyExists {});
-    /// # Ok(())
-    /// # })
-    /// # }
+    /// You can also use the [send_request](CacheClient::send_request) method to create a cache using a [CreateCacheRequest].
```
cprice404, line 73: if we are going to move examples out of these docs, we need to make sure we make it really obvious that the send_request path provides optional arguments that are not available in the shorthand API. In addition to adding verbiage to that effect I almost think that we should enumerate the optional arguments from the request object here. It's important that users know why they would want to opt in to that alternate syntax.

@malandis any thoughts?

src/cache_client.rs
```diff
@@ -167,41 +143,35 @@ impl CacheClient {
...
+    /// If you use [send_request](CacheClient::send_request) to set an item using a
+    /// [SetRequest], you can also provide the following optional arguments:
     ///
-    /// ```
-    /// # fn main() -> anyhow::Result<()> {
-    /// # use momento_test_util::create_doctest_cache_client;
-    /// # tokio_test::block_on(async {
-    /// # let (cache_client, cache_name) = create_doctest_cache_client();
-    /// cache_client.set(&cache_name, "k1", "v1").await?;
+    /// * `ttl` - The time-to-live for the item. If not provided, the client's default time-to-live is used.
```
anitarua, line 150: @cprice404 
All APIs with optional args except for `set` had an "Optional Arguments" section that I kept. Should I repeat the same verbiage on that last line? 

As for the APIs without additional optional arguments, I linked to the request objects anyway in case they want to send request objects directly but wasn't sure what else to mention there

src/requests/cache/sorted_set/sorted_set_fetch_by_rank.rs
```diff
@@ -61,6 +61,7 @@ pub enum SortOrder {
 /// # Ok(())
 /// # })
 /// # }
+/// ```
```
cprice404, line 64: lol wow :)

cprice404 (commented):
thank you so much for doing this! let's get aligned on how to make sure that users know when/why to use requests vs. the cacheclient fns but otherwise lgtm!

src/cache_client.rs
malandis: Agreed. If we presently have optional arguments and building the request is the way to do it, that should be in the examples. If at present there are no optional arguments, then it's less important, but if we change the request at some point to add optional arguments, we should demonstrate it in the docs.

anitarua:
@cprice404 @malandis 

Docs example for Set would look like this on the CacheClient, with the Optional Arguments heading above the shorthand example and the `SetRequest` and `optional arguments` links below the example:
Screenshot 2024-04-01 at 9 52 40 AM

The `SetRequest` and `optional arguments` links will direct them to the SetRequest docs page with the example of using `send_request`:
Screenshot 2024-04-01 at 9 54 23 AM

I applied this pattern to all the APIs with optional arguments so far. Thoughts?

malandis:
Looks really nice!

cprice404 (approved):
I reviewed a sampling of the requests where there were optional things available and I think your verbiage sounds great. Thanks for your patience on this one 🚢 

thank you

PR #214, opened 2024-03-28
chore: make build fail when there are docs warnings

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/199

Adds new Makefile target to build the docs with a flag set to fail when there are warnings. 
Adds new `on-pull-request` workflow step to build the docs with that Makefile target.

First commit should fail with docs warning, second commit should fix the issue and the build should pass.


malandis (commented):
Very nice!

malandis (approved):

Issue #192, opened 2024-03-21
Decide on default arguments for sorted set fetch by rank and score


nand4011:
The sorted_set_fetch_by_rank and sorted_set_fetch_by_score functions in the new cache client both take cache name, sorted set name, and sort order, and they both return the same response object. i.e. they are the same function. We need to figure out which additional arguments we should add to either of them to separate them.
Idea:
sorted_set_fetch that takes cache name, sorted set name, sort order, limit, offset
sorted_set_fetch_by_rank that takes cache name, sorted set name, sort order, start_rank, end_rank
sorted_set_fetch_by_score that takes cache name, sorted set name, sort order, min_score, max_score

anitarua:
Not actually sure if any action needs to be taken here. Fetch by score and rank appear to have optional arguments in all the SDKs they're implemented in:

* JS fetch by [rank](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/internal/clients/cache/AbstractCacheClient.ts#L1255) and [score](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/internal/clients/cache/AbstractCacheClient.ts#L1293)
* Python fetch by [rank](https://github.com/momentohq/client-sdk-python/blob/79bc60ef75ea098112403f91b11f2b221be58b14/src/momento/cache_client.py#L927) and [score](https://github.com/momentohq/client-sdk-python/blob/79bc60ef75ea098112403f91b11f2b221be58b14/src/momento/cache_client.py#L894)
* Java fetch by [rank](https://github.com/momentohq/client-sdk-java/blob/660519a415a18ef2b0b1ca798231abbcc042cc65/momento-sdk/src/main/java/momento/sdk/CacheClient.java#L1036) and [score](https://github.com/momentohq/client-sdk-java/blob/660519a415a18ef2b0b1ca798231abbcc042cc65/momento-sdk/src/main/java/momento/sdk/CacheClient.java#L1077)
* Go fetch by [rank](https://github.com/momentohq/client-sdk-go/blob/7858197b97c4574407c5eb9caa013bc389760975/momento/sorted_set_fetch_by_rank.go#L11) and [score](https://github.com/momentohq/client-sdk-go/blob/7858197b97c4574407c5eb9caa013bc389760975/momento/sorted_set_fetch_by_score.go#L11)
[...]

PR #215, opened 2024-03-29
feat: add start_rank and end_rank as optional arguments to SortedSetFetchByRank

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/192 

As discussed with Chris and Michael, it's pretty likely users will call fetch by rank to paginate through their sorted set, whereas we can't easily predict what arguments they'll want to set for fetch by score. Hence we're adding the start_rank and end_rank optional arguments to the fetch by rank API as required arguments to differentiate it from fetch by score.

### DIFF
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -556,7 +556,9 @@ impl CacheClient {
     /// let fetch_response = cache_client.sorted_set_fetch_by_rank(
     ///     cache_name,
     ///     sorted_set_name,
-    ///     SortOrder::Ascending
+    ///     SortOrder::Ascending,
+    ///     None,
+    ///     None
     /// ).await?;
     ///
     /// match fetch_response {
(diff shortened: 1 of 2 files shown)


cprice404 (approved):
lgtm! @nand4011 after discussing how to differentiate the two functions this is what we ended up aligning on. let us know if you have thoughts.

nand4011 (commented):
Are these the only fetch arguments we want to expose? I could see a sorted_set_fetch that gets the whole thing in addition to a fetch_by_score with min/max scores and this, or a fetch that exposes offset and count.

anitarua:
@nand4011 wouldn't adding a new `sorted_set_fetch` API break consistency with the other SDKs?
Since there's no overloading, other variations of fetch by score would need to be named something different too (e.g. `fetch_by_score_with_offset_and_count`), and that would also be different from the other SDKs.

malandis (approved):

Issue #164, opened 2024-02-15
Add unimplemented scalar cache methods


nand4011:
The rust sdk does not currently support these scalar methods:

- [x] keyExists, keysExist
- [x] itemGetType, delete, increment
- [x] updateTtl, increaseTtl, decreaseTtl, itemGetTtl
- [x] setIfAbsent, setIf*

pratik151192:
@pratik151192 picking `setIfNotExists`

PR #217, opened 2024-04-08
feat: add keyExists and keysExist

anitarua:
Addresses part of https://github.com/momentohq/client-sdk-rust/issues/164

### DIFF
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -606,10 +606,9 @@ impl CacheClient {
     /// use momento::requests::cache::scalar::keys_exist::KeysExist;
     ///
     /// let result = cache_client.keys_exist(&cache_name, vec!["key1", "key2", "key3"]).await?;
-    /// if !result.exists.is_empty() {
-    ///     println!("Processing list of booleans: {:#?}", result.exists);
-    /// } else {
-    ///     println!("Received empty list of booleans!");
+    /// println!("Expecting all keys to exist:");
+    /// for (key, exists) in result.exists_dictionary() {
+    ///    println!("Key: {}, Exists: {}", key, exists);
     /// }
     /// # Ok(())
     /// # })
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -606,9 +606,13 @@ impl CacheClient {
     /// use momento::requests::cache::scalar::keys_exist::KeysExist;
     /// use std::collections::HashMap;
     ///
-    /// let result = cache_client.keys_exist(&cache_name, vec!["key1", "key2", "key3"]).await?;
-    /// let exists_map: HashMap<String, bool> = result.into();
-    /// println!("Expecting all keys to exist: {:#?}", exists_map);
+    /// // Receive results as a HashMap
+    /// let result_map: HashMap<String, bool> = cache_client.keys_exist(&cache_name, vec!["key1", "key2", "key3"]).await?.into();
+    /// println!("Expecting all keys to exist: {:#?}", result_map);
+    ///
+    /// // Or receive results as a Vec
+    /// let result_list: Vec<bool> = cache_client.keys_exist(&cache_name, vec!["key1", "key2", "key3"]).await?.into();
+    /// println!("Expecting all keys to exist: {:#?}", result_list);
     /// # Ok(())
     /// # })
     /// # }
--- a/src/cache_client.rs
+++ b/src/cache_client.rs
@@ -14,6 +14,8 @@ use crate::requests::cache::create_cache::{CreateCache, CreateCacheRequest};
 use crate::requests::cache::delete_cache::{DeleteCache, DeleteCacheRequest};
 use crate::requests::cache::flush_cache::{FlushCache, FlushCacheRequest};
 use crate::requests::cache::list_caches::{ListCaches, ListCachesRequest};
+use crate::requests::cache::scalar::key_exists::{KeyExists, KeyExistsRequest};
+use crate::requests::cache::scalar::keys_exist::{KeysExist, KeysExistRequest};
 use crate::requests::cache::set::set_add_elements::{SetAddElements, SetAddElementsRequest};
 use crate::requests::cache::sorted_set::sorted_set_fetch_by_rank::{
     SortOrder, SortedSetFetchByRankRequest,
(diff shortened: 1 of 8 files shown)


anitarua:
TIL:
* `assert_eq` should be used for matching strings instead of `let _err_msg = "not found".to_string(); assert!(matches!(result.to_string(), _err_message));` because the `_` in front of the variables mean we're intentionally not using those variables, so the asserts weren't working as we actually expected. Will need to audit the existing cache control client and sorted set tests to make sure assertions are actually doing their job when checking for matching strings and enum cases
* `assert!()` accepts informative `panic` messages: `assert!(
        result.exists.is_empty(),
        "Expected empty list of bools given no keys, but received {:#?}",
        result.exists
    );`

src/requests/errors.rs
```diff
@@ -4,7 +4,7 @@ use tonic::codegen::http;
 
 use crate::auth::AuthError;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
anitarua, line 7: allows us to use `assert_eq()` to check for expected Momento error codes

cprice404:
> TIL:
> 
>     * `assert_eq` should be used for matching strings instead of `let _err_msg = "not found".to_string(); assert!(matches!(result.to_string(), _err_message));` because the `_` in front of the variables mean we're intentionally not using those variables, so the asserts weren't working as we actually expected. Will need to audit the existing cache control client and sorted set tests to make sure assertions are actually doing their job when checking for matching strings and enum cases
> 
>     * `assert!()` accepts informative `panic` messages: `assert!( result.exists.is_empty(), "Expected empty list of bools given no keys, but received {:#?}", result.exists );`

2 very nice finds, thank you!

src/requests/cache/mod.rs
```diff
@@ -7,6 +7,7 @@ pub mod flush_cache;
 pub mod list_caches;
 
 pub mod basic;
+pub mod scalar;
```
cprice404, line 10: nit - I put the `Get` request and friends in `basic` and I see you're adding `scalar` here. I don't have a strong pref which of those we go with but I'd probably just stick with one or the other. It's kinda the "everything except collections" directory IMO. but if you think they should be separated lmk!

src/requests/cache/basic/get.rs
```diff
@@ -58,8 +58,9 @@ impl<K: IntoBytes> MomentoRequest for GetRequest<K> {
     type Response = Get;
 
     async fn send(self, cache_client: &CacheClient) -> MomentoResult<Get> {
-        let request = prep_request(
+        let request = prep_request_with_timeout(
```
cprice404, line 61: oh this is interesting.

I'm not sure we should even *have* a `prep_request` function, it kinda seems like we should *only* have the `_with_timeout` one. WDYT? (not an issue for this PR but something we may want to come back and clean up.)

src/cache_client.rs
```diff
@@ -552,6 +554,77 @@ impl CacheClient {
...
+    /// ```
+    /// # fn main() -> anyhow::Result<()> {
+    /// # use momento_test_util::create_doctest_cache_client;
+    /// # tokio_test::block_on(async {
+    /// # let (cache_client, cache_name) = create_doctest_cache_client();
+    /// use momento::requests::cache::scalar::keys_exist::KeysExist;
+    ///
+    /// let result = cache_client.keys_exist(&cache_name, vec!["key1", "key2", "key3"]).await?;
+    /// if !result.exists.is_empty() {
+    ///     println!("Processing list of booleans: {:#?}", result.exists);
```
cprice404, line 610: I suspect that as a user the main way I'd prefer to interact with this is to have basically a "HashMap" type of interface available on the result. So I can do like, `result.hash_map.get("key1")` or something like that. The list of booleans is necessary as a lowest-level fallback, but generally as a user I'd prefer to avoid it. Maybe worth a quick look to see if that would be easy to add? I think we have a few examples of this in other SDKs.

cprice404 (commented):
looking good! just a few minor thoughts/questions.

src/requests/cache/basic/get.rs
anitarua: definitely!

src/requests/cache/mod.rs
anitarua: I think "scalar" is more in line with the other SDKs. I can come back to clean this up when doing the cleaning-up-exports ticket

src/cache_client.rs
anitarua: Looks like it's inconsistent across SDKs. I agree that would be a good UX, can definitely add a hash map here and where they're currently missing:

**Provides hash map**
* [Dotnet sdk](https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/Responses/CacheKeysExistResponse.cs)
* [PHP sdk](https://github.com/momentohq/client-sdk-php/blob/4fb88057f3a4cdad2f186c811043c171e1d10ce4/src/Cache/CacheOperationTypes/CacheOperationTypes.php#L862)

**Missing hash map**
* [JS sdk](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/messages/responses/cache-keys-exist.ts)
* [Go sdk](https://github.com/momentohq/client-sdk-go/blob/main/responses/keys_exist.go)

cprice404: oh wow good catch. can you create a ticket for adding that to the JS SDK? :)

probably the dotnet sdk is a good one to use for inspiration for now though. feel free to timebox.

src/requests/cache/scalar/keys_exist.rs
```diff
@@ -0,0 +1,106 @@
...
+                .zip(response.exists.clone())
+                .collect(),
+        })
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct KeysExist {
+    pub exists: Vec<bool>,
+    pub exists_dictionary: HashMap<String, bool>,
```
anitarua, line 95: was unable to get HashMap<IntoBytes, bool> working, but got a String version working. Can probably add Vec<u8> version as well if that seems useful

cprice404: I think string version is great for now, thx

cprice404: rather than exposing this as another property on the struct, could we expose it via an `into`/`From` thing? and show it in the docstring example? I think you will need to save the keys into the struct in order to support that, but it could be a private property rather than a public one?

Just thinking about trying to make the patterns for accessing these different shapes of the data consistent from one API to the next, and e.g. for Get, we mostly expose alternate types via `into`. WDYT?

cprice404 (commented):
just the one question about the into

src/requests/cache/scalar/keys_exist.rs
anitarua: that makes sense, made those fields private and implemented `From` in https://github.com/momentohq/client-sdk-rust/pull/217/commits/11c6c6d7092def3089a3a5b0c8bd12d7a31b3b68 

should I keep the `exists` and `exists_dictionary` method? or remove those?

cprice404: I think for now let's remove them and just show examples of the `into`s. Always easier to add more API surface area later than to remove it! :)

cprice404 (commented):
let's remove the fn versions and just document the `into` versions for now. then I am 👍 on this

src/requests/cache/scalar/keys_exist.rs
```diff
@@ -21,17 +21,16 @@ use crate::{CacheClient, IntoBytes, MomentoResult};
...
 ///     cache_name,
 ///     vec!["key1", "key2", "key3"]
 /// );
 ///
 /// let result = cache_client.send_request(request).await?;
-/// println!("Expecting all keys to exist:");
-/// for (key, exists) in result.exists_dictionary() {
-///    println!("Key: {}, Exists: {}", key, exists);
-/// }
+/// let exists_map: HashMap<String, bool> = result.into();
```
cprice404, line 32: do you want to include an example of the Vec into as well?

cprice404 (approved):
whoohoo

PR #218, opened 2024-04-10
chore: refactor tests for better consistency, more accurate asserts

anitarua:
Trying to revise existing tests to better align what we want for best practices:
* avoid using `unwrap`, use `?` instead as per the [contributing docs](https://github.com/momentohq/client-sdk-rust/blob/main/CONTRIBUTING.md)
* tests should return a `MomentoResult` (see [this comment](https://github.com/momentohq/client-sdk-rust/pull/209#discussion_r1542177360))
* avoid `clone()`-ing things unnecessarily
* ran into an issue with `assert!(matches!(...))` not actually working as expected with strings (see [this comment](https://github.com/momentohq/client-sdk-rust/pull/217#issuecomment-2043730017)). Preferring to compare `MomentoErrorCode`s instead when expecting errors

### DIFF
--- a/tests/cache_sorted_set.rs
+++ b/tests/cache_sorted_set.rs
@@ -20,13 +20,7 @@ async fn sorted_set_put_element_happy_path() -> MomentoResult<()> {
     let score = 1.0;
 
     let result = client
-        .sorted_set_fetch_by_rank(
-            cache_name,
-            &*sorted_set_name,
-            Ascending,
-            None,
-            None,
-        )
+        .sorted_set_fetch_by_rank(cache_name, &*sorted_set_name, Ascending, None, None)
         .await?;
     assert_eq!(result, SortedSetFetch::Miss);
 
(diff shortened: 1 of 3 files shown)


tests/cache_sorted_set.rs
```diff
@@ -10,40 +12,26 @@ use momento::requests::cache::sorted_set::sorted_set_fetch_response::SortedSetFe
...
-        )
-        .await
-        .unwrap();
+        .sorted_set_fetch_by_rank(cache_name, &*sorted_set_name, Ascending, None, None)
+        .await?;
     assert_eq!(result, SortedSetFetch::Miss);
 
     client
-        .sorted_set_put_element(cache_name.clone(), sorted_set_name.clone(), "value", 1.0)
+        .sorted_set_put_element(cache_name, &*sorted_set_name, "value", 1.0)
```
cprice404, line 28: not relevant to this PR but I really want to understand this; why is the `&*` required, why doesn't just `&` work? would you mind capturing that in a separate ticket?

cprice404 (approved):
this is a really nice cleanup, thanks a bunch for doing it

tests/cache_sorted_set.rs
anitarua: I think it's to dereference it to get the underlying type, then create a reference from that underlying type: https://stackoverflow.com/questions/41273041/what-does-combined-together-do-in-rust

PR #219, opened 2024-04-11
feat: add Delete, Increment, ItemGetType

anitarua:
Following the KeyExists and KeysExist PR, this one adds more scalar APIs: Delete, Increment, and ItemGetType


src/cache/requests/scalar/item_get_type.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    fn try_from(value: ItemGetType) -> Result<Self, Self::Error> {
+        match value {
+            ItemGetType::Hit { key_type } => Ok(key_type),
+            ItemGetType::Miss => Err(MomentoError {
+                message: "item get type response was a miss".into(),
+                error_code: MomentoErrorCode::Miss,
+                inner_error: None,
+                details: None,
+            }),
```
anitarua, line 149: tried to make this match the `get` response `into`/`from` accessors as much as possible

tests/cache_scalar.rs
```diff
@@ -108,3 +112,168 @@ async fn keys_exist_happy_path() -> MomentoResult<()> {
...
+    match result {
+        ItemGetType::Hit { key_type } => assert_eq!(
+            key_type,
+            ItemType::SortedSet,
+            "Expected SortedSet, got {:?}",
+            key_type
+        ),
+        _ => panic!("Expected Hit, got {:?}", result),
+    }
+    client.delete(cache_name, &**key).await?;
```
cprice404, line 223: oh this is interesting, we should make sure we remember to come back and add Dictionary/List here once those exist.

cprice404 (approved):
:ty:

Issue #194, opened 2024-03-21
decide how we want to organize the exports


cprice404:

cprice404:
Michael, Anita, Nate and I had a meeting to discuss this.

The plan of record for now is to have the exported module names be, e.g., `momento::cache::` and `momento::topic::`, etc. All of the relevant types for each product will live in the root of these namespaces, e.g. `momento::cache::{GetRequest,Get}`, etc.

Whomever picks up this ticket can take a pass at making this change for a handful of types and then solicit feedback before we go all the way.

Also of note for this - we want to get rid of the exports for the `proto` types that are currently exported.

PR #221, opened 2024-04-16
feat: reorganize exports

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/194 

* Removed the proto exports from the top-level `lib.rs`; re-exported out of the `response` namespace instead.
* Renamed `basic` folder to `scalar` to be more consistent with other SDKs' naming schemes.
* Renamed the `requests` folder to `cache` and the `cache` folder to `requests`.
* Re-exported stuff individually out of `requests` (thus hiding `requests` from the import path and putting it under the `cache` namespace instead).


example/src/bin/readme.rs
```diff
@@ -1,5 +1,5 @@
 use momento::config::configurations::laptop;
-use momento::requests::cache::basic::get::Get;
+use momento::cache::{Get, SortOrder, SortedSetFetch};
```
anitarua, line 2: simplified import path

src/cache/mod.rs
```diff
@@ -0,0 +1,25 @@
+// Make module "requests" private and re-export its contents so users can
+// import using "momento::cache::{Get, Set, ...}"
+mod requests;
```
anitarua, line 3: Re-exporting as per https://doc.rust-lang.org/rustdoc/write-documentation/re-exports.html

src/lib.rs
```diff
@@ -1,9 +1,15 @@
...
+*/
 pub mod auth;
+pub mod cache;
 /// Configuration structs for the Momento clients.
 pub mod config;
+pub mod errors;
 pub mod preview;
-pub mod requests;
 pub mod response;
+/*************************************************************/
```
anitarua, line 12: are we okay with these namespaces?

hadn't rearranged anything related to topics yet, but should we rename it from "preview" to just "topics"?

example/src/bin/readme.rs
```diff
@@ -17,6 +17,13 @@ pub async fn main() -> Result<(), MomentoError> {
 
     cache_client.create_cache(CACHE_NAME.to_string()).await?;
 
+    cache_client.flush_cache(CACHE_NAME).await?;
+
+    match cache_client.list_caches().await {
```
cprice404, line 22: FYI - this readme.ts file gets sucked into the README template, so we want to keep it super short. If you want to make a copy of this file and call it `cache_operations.rs` or something, and keep the `readme.ts` file short like it was before, that would work!

example/src/bin/readme.rs
cprice404: ❤️

src/cache/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+pub use requests::delete_cache::{DeleteCache, DeleteCacheRequest};
+pub use requests::flush_cache::{FlushCache, FlushCacheRequest};
+pub use requests::list_caches::{ListCaches, ListCachesRequest};
+
+pub use requests::scalar::get::{Get, GetRequest, GetValue};
+pub use requests::scalar::set::{Set, SetRequest};
+
+pub use requests::set::set_add_elements::{SetAddElements, SetAddElementsRequest};
+
+pub use requests::sorted_set::sorted_set_fetch_by_rank::{SortOrder, SortedSetFetchByRankRequest};
```
cprice404, line 17: if this SortOrder is really specific to SortedSets we might need to disambiguate that in some way, like name it SortedSetSortOrder or SortedSet::SortOrder or something. If it's not specific to sorted sets then we could move its definition to a more generic location. Just a nit; this is not really something specific to this PR, can sort it out later (pun intended)

src/lib.rs
cprice404: good callout, ty

let's make another pass to deal with topics after we are happy with cache stuff.

response namespace will go away once we remove the old client, i think.

we should check out auth and config and see if we want to re-export stuff from them either directly under `momento`, or into `cache` (e.g. for configuration, which is probably going to be cache-specific). Don't feel obligated to hit those in this PR, but maybe make a ticket for it.

src/lib.rs
```diff
@@ -13,19 +19,12 @@ mod grpc;
...
 pub use crate::cache_client::CacheClient;
 
 pub type MomentoResult<T> = Result<T, MomentoError>;
-
-pub mod sorted_set {
-    pub use momento_protos::cache_client::sorted_set_fetch_request::{Order, Range};
-    pub use momento_protos::cache_client::sorted_set_fetch_response::found::Elements;
-    pub use momento_protos::cache_client::sorted_set_fetch_response::SortedSet;
-    pub use momento_protos::cache_client::SortedSetElement;
-}
```
cprice404, line 31: :ty:

cprice404 (commented):
left a few comments but I don't think any of them are blockers except for keeping `readme.ts` short. looking good!

example/src/bin/readme.rs
anitarua: was planning on reverting the changes in `readme.ts` before taking this PR out of draft, just committed it for review purposes for now 👍

cprice404: gotcha. very helpful for review ty

src/lib.rs
anitarua: captured in ticket https://github.com/momentohq/client-sdk-rust/issues/222

src/cache/mod.rs
anitarua: Looks like it's called SortedSetOrder in the other SDKs so will follow suit here

src/response/mod.rs
```diff
@@ -29,6 +29,13 @@ pub use self::generate_api_token_response::*;
 pub use self::list_cache_response::*;
 pub use self::list_signing_keys_response::*;
 
+pub mod simple_cache_client_sorted_set {
+    pub use momento_protos::cache_client::sorted_set_fetch_request::{Order, Range};
+    pub use momento_protos::cache_client::sorted_set_fetch_response::found::Elements;
+    pub use momento_protos::cache_client::sorted_set_fetch_response::SortedSet;
+    pub use momento_protos::cache_client::SortedSetElement;
+}
+
```
anitarua, line 38: these were still used in the simple cache client doctests so I opted to re-export them from here to make it easier to remove these once we remove the simple cache client entirely

cprice404 (dismissed):
yeah

cprice404 (approved):

Issue #222, opened 2024-04-16
re-export additional items?


anitarua:
See https://github.com/momentohq/client-sdk-rust/pull/221#discussion_r1567921033

* Move topics out of "preview" namespace?
* Re-export items from "auth" or "config" so that they're exported from "cache" or "momento" namespace instead?

PR #224, opened 2024-04-18
feat: reorganize topics exports

anitarua:
Addresses #222 

- Renamed and re-exported topics client. Created a "config" folder under the "topics" folder for future topics configs.
- Leaving "auth" namespace as is since `CredentialProvider`s are used across cache and topics
- Moved existing cache config objects under the "cache" namespace.


cprice404 (dismissed):

scripts/test-teardown.rs
```diff
@@ -1,6 +1,6 @@
 use std::time::Duration;
 
-use momento::config::configurations;
+use momento::cache::configurations;
```
cprice404, line 3: ❤️

cprice404 (approved):
heart

PR #225, opened 2024-04-20
chore: minor test cleanups

cprice404:
This commit does two things:

1. Adds a test utility function to DRY up all the places where
   we were creating UUIDs to create unique strings, and
2. In all the places where the tests were doing `&*` or `&**` to
   get access to a `&str` to pass to the APIs under test, we now
   just create two variables. The first variable is a `String`,
   so that the lifetime / ownership of the `String` extends through
   the end of the function; the second variable is a `&str` acquired
   via `&str`. This can be passed to the APIs under test as many times
   as possible without a "move" causing a compile failure, and prevents
   the need to do `.clone()` or `.as_str()` at every call site.

### DIFF
--- a/test-util/src/lib.rs
+++ b/test-util/src/lib.rs
@@ -4,5 +4,5 @@ mod test_utils;
 pub use crate::cache_test_state::CACHE_TEST_STATE;
 pub use crate::test_utils::{
     create_doctest_cache_client, doctest, get_test_cache_name, get_test_credential_provider,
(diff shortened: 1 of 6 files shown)


cprice404:
@anitarua let me know what you think of this. I'm not super attached to it, was just playing around to see if I could find something that felt a little more natural to me.

anitarua (approved):
Looks great to me! Thanks for doing this cleanup!

PR #228, opened 2024-04-23
feat: re-export CollectionTtl under cache namespace

anitarua:
another reorganization to move `CollectionTtl` out of `simple_cache_client` and changes the import path from `momento::CollectionTtl` to `momento::cache::CollectionTtl` since it's only used for cache collection data types


src/cache/collection_ttl.rs
```diff
@@ -0,0 +1,92 @@
...
+///
+/// The default behaviour is to refresh the TTL (to prolong the life of the collection)
+/// each time it is written using the client's default item TTL.
+#[derive(Copy, Clone, Debug)]
+pub struct CollectionTtl {
+    ttl: Option<Duration>,
+    refresh: bool,
+}
+
+impl CollectionTtl {
```
cprice404, line 20: while you are in here can you do a quick spot check on the dev docs / javascript version of this class and add any convenience factory fns that we are missing? e.g. I am pretty sure we document `CollectionTtl.of(...)` as the most common factory fn in JS.

malandis (approved):

PR #229, opened 2024-04-23
chore: update client protos dependency

anitarua:
We'll need a newer version of the client protos dependency to implement the `SetIf*` APIs


cprice404 (approved):
approve

PR #233, opened 2024-04-23
test: add test data types to ease test data creation

malandis:
A common concern in the integration tests is creating test data. In
Rust, creating test data is a bit more painful than in other
languages. Thus we introduce some helper structs to create test data.

These test data create random data for each data type we have. In the
future we can also add comparators (PartialEq implementations) which
will be a common concern and pain in the test code.

### DIFF
--- a/test-util/src/test_data.rs
+++ b/test-util/src/test_data.rs
@@ -1,0 +1,101 @@
+use uuid::Uuid;
+
+pub fn unique_string(prefix: impl Into<String>) -> String {
+    format!("{}-{}", prefix.into(), Uuid::new_v4())
+}
+
+pub fn unique_cache_name() -> String {
+    unique_string("rust-sdk")
+}
+
+pub fn unique_key() -> String {
+    unique_string("key")
+}
+
+pub fn unique_value() -> String {
+    unique_string("value")
+}
+
+pub struct TestScalar {
+    pub key: String,
+    pub value: String,
+}
+
+impl TestScalar {
+    pub fn new() -> Self {
+        Self {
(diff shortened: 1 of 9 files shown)


test-util/src/test_data.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+pub fn unique_value() -> String {
+    unique_string("value")
+}
+
+pub struct TestScalar {
+    pub key: String,
+    pub value: String,
+}
```
malandis, line 22: We may need to generalize these over the `IntoBytes` type to test both bytes and strings in a coming PR.

test-util/src/test_data.rs
```diff
@@ -0,0 +1,101 @@
...
+            value: unique_value(),
+        }
+    }
+
+    pub fn key(&self) -> String {
+        self.key.clone()
+    }
+
+    pub fn value(&self) -> String {
+        self.value.clone()
```
anitarua, line 37: wouldn't this mean we'd be cloning strings a lot more often instead of using references?

malandis: Yes it would. Though this isn't performance critical, that's a better thing to do as a matter of practice. Thanks! Changed in [82ae3cc](https://github.com/momentohq/client-sdk-rust/pull/233/commits/82ae3ccbbe8adc988614392773ef7bb8aec727ca) and [512557f](https://github.com/momentohq/client-sdk-rust/pull/233/commits/512557f5d77982acec2af3f4388105216840ff7a)

anitarua (approved):
lgtm!

PR #234, opened 2024-04-24
feat: add SetIf* methods

anitarua:
Last part of #164 
Adds the SetIf* methods (except for the deprecated SetIfNotExists):

- SetIfAbsent
- SetIfPresent
- SetIfEqual
- SetIfNotEqual
- SetIfPresentAndNotEqual
- SetIfAbsentOrEqual

### DIFF
--- a/tests/cache/scalar.rs
+++ b/tests/cache/scalar.rs
@@ -241,7 +241,7 @@ mod set_if_absent {
     }
 
     #[tokio::test]
-    async fn string_key_string_value() -> MomentoResult<()> {
+    async fn happy_path() -> MomentoResult<()> {
         let client = &CACHE_TEST_STATE.client;
         let cache_name = CACHE_TEST_STATE.cache_name.as_str();
         let item1 = TestScalar::new();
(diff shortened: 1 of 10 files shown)


malandis:
This looks good. Would you be able to refactor the tests to use the TestScalar object I created? That should ease some of the boilerplate and make the pattern consistent.

tests/cache/scalar.rs
```diff
@@ -208,18 +214,1121 @@ mod increment {
...
+        let cache_name = CACHE_TEST_STATE.cache_name.as_str();
+        let item1 = TestScalar::new();
+        let item2 = TestScalar::new();
+
+        // Setting a key that doesn't exist should create it
+        let result = client
+            .set_if_absent(cache_name, item1.key(), item1.value())
+            .await?;
+        assert_eq!(result, SetIfAbsent::Stored);
+        let result_value: String = client.get(cache_name, item1.key()).await?.try_into()?;
```
malandis, line 255: This is a good pattern. We can use this or the `Get::from` that I demonstrated. For this I recommend adding a message if the `try_into` fails:

```suggestion
        let result = client.get(cache_name, item1.key()).await?;
        let result_value: String = result.try_into().expect("Expected Get::Hit but got {?:}", result);
```

That way our integration tests can have something to say about what the value was whenever the `try_into` fails.

tests/cache/scalar.rs
```diff
@@ -208,18 +214,1121 @@ mod increment {
...
+        let result_value: Vec<u8> = client.get(cache_name, item1.key()).await?.try_into()?;
+        assert_eq!(result_value, item1.value().as_bytes());
+
+        // Setting a key that exists should not overwrite it
+        let result = client
+            .set_if_absent(cache_name, item1.key(), item2.value())
+            .await?;
+        assert_eq!(result, SetIfAbsent::NotStored);
+        let result_value: Vec<u8> = client.get(cache_name, item1.key()).await?.try_into()?;
+        assert_eq!(result_value, item1.value().as_bytes());
```
malandis, line 290: I understand the intent here. In other repos where we exhaustively tested input types, we tested passing a bytes key and string value, bytes key and bytes value, etc.

To contrast here the data in `TestScalar` is always `String` based, so we are really storing string data, getting the string content, converting to bytes and comparing based on bytes. This isn't quite the same thing as originally storing bytes and then comparing bytes on the way out.

For now based on our discussion this morning, we do not need such exhaustive coverage. You can limit the tests to String-only for now. I appreciate the thorough eye for tests though!

malandis (commented):
Really appreciate the thorough documentation and test coverage 👍

tests/cache/scalar.rs
anitarua: I like the `Get::from` approach a lot, and the `assert_eq!` should handle showing the diff if the result isn't as expected, so I'll switch similar instances to `Get::from`

malandis (approved):
✨

PR #235, opened 2024-04-24
test: more scalar coverage and test conveniences

malandis:
We audit test coverage for `get`, `set`, and `sorted sets` and add tests as needed. We also implement traits and conveniences that will make future integration test authorship easier. In particular:

- For each test data type, we implement the `From` trait paired with the corresponding cache hit response. That way we can use the test data directly in an assertion without the need to massage it in the right form.
- We extract asserting a collection fetch is equal to test data. Both unpacking the actual result, massaging the expected reuslt, and comparing contain a lot of boilerplate. The sorted set tests demonstrate this pattern.

### DIFF
--- a/test-util/src/test_data.rs
+++ b/test-util/src/test_data.rs
@@ -1,5 +1,5 @@
 use momento::{
-    cache::{Get, GetValue},
+    cache::{Get, GetValue, SortedSetElements, SortedSetFetch},
     IntoBytes,
 };
 use uuid::Uuid;
@@ -20,6 +20,7 @@ pub fn unique_value() -> String {
     unique_string("value")
 }
 
+#[derive(Debug, PartialEq, Clone)]
 pub struct TestScalar {
     pub key: String,
     pub value: String,
@@ -40,13 +41,6 @@ impl TestScalar {
     pub fn value(&self) -> &str {
         &self.value
     }
-
-    // todo use `into` after getting to sorted sets
-    pub fn to_get_hit(&self) -> Get {
-        Get::Hit {
-            value: GetValue::new(self.value().into_bytes()),
-        }
-    }
 }
 
 impl Default for TestScalar {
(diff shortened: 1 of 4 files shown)


tests/cache/scalar.rs
```diff
@@ -1,9 +1,97 @@
...
+            item.key(),
+            cache_name,
+            result
+        );
+
+        // Getting the key should return a hit with the value
+        let result = client.get(cache_name, item.key()).await?;
+        assert_eq!(
+            result,
+            Get::from(&item),
```
malandis, line 44: Here we use `From` to easily compare the test data to retrieved data.

tests/cache/sorted_set.rs
```diff
@@ -1,78 +1,139 @@
...
+            result,
+            vec![
+                ("1".to_string(), 0.0),
+                ("3".to_string(), 0.5),
+                ("2".to_string(), 1.0),
+                ("5".to_string(), 1.5),
+                ("4".to_string(), 2.0),
+            ],
+        )
+        .await?;
```
malandis, line 96: The tests are much easier to follow after adding `From` and the helper function.

test-util/src/test_data.rs
```diff
@@ -44,6 +49,15 @@ impl Default for TestScalar {
     }
 }
 
+impl From<&TestScalar> for Get {
+    fn from(test_scalar: &TestScalar) -> Self {
+        Get::Hit {
+            value: GetValue::new(test_scalar.value().into_bytes()),
+        }
+    }
+}
```
malandis, line 58: This allows us to write:

```rust
let expected = TestScalar { key: unique_key(), value: unique_value() };
// ...
let actual = cache_client.get(cache_name, expected.key());
assert_eq!(actual, Get::from(expected));
```

anitarua (approved):

Issue #165, opened 2024-02-15
Add unimplemented list cache methods


nand4011:
The Rust SDK currently does not support these list methods:

- listConcatenateBack
- listConcatenateFront
- listRetain

anitarua:
New methods:

- listConcatenateBack
- listConcatenateFront
- listRetain

Port from `simple_cache_client` to `cache_client`:

- listFetch
- listLength
- listPopBack
- listPopFront
- listPushBack
- listPushFront
- listRemoveValue

PR #237, opened 2024-04-29
feat: add list collection

anitarua:
Addresses #165 

Prioritized getting these APIs implemented:

- listConcatenateBack
- listConcatenateFront
- listFetch
- listLength
- listPopBack
- listPopFront
- listRemoveValue

listPushFront and listPushBack were also implemented in the simple_cache_client but seemed less urgent to include them here since the same functionality can be done using the concatenate methods.

### DIFF
--- a/src/cache/requests/list/list_concatenate_back.rs
+++ b/src/cache/requests/list/list_concatenate_back.rs
@@ -6,6 +6,9 @@ use crate::{
 
 /// Adds multiple elements to the back of the given list. Creates the list if it does not already exist.
 ///
+/// Appends the supplied list to the end of an existing list. For example, if you have a list [1, 2, 3]
+/// and listConcatenateBack [4, 5, 6], you will create [1, 2, 3, 4, 5, 6].
+///
 /// # Arguments
 /// * `cache_name` - name of cache
 /// * `list_name` - name of the list
--- a/src/cache/requests/list/list_concatenate_back.rs
+++ b/src/cache/requests/list/list_concatenate_back.rs
@@ -1,0 +1,100 @@
+use crate::{
+    cache::{CollectionTtl, MomentoRequest},
+    utils::prep_request_with_timeout,
+    CacheClient, IntoBytes, MomentoResult,
+};
+
+/// Adds multiple elements to the back of the given list. Creates the list if it does not already exist.
+///
+/// # Arguments
+/// * `cache_name` - name of cache
+/// * `list_name` - name of the list
+/// * `values` - list of values to add to the back of the list
+///
+/// # Optional Arguments
+///
+/// * `collection_ttl` - The time-to-live for the collection. If not provided, the client's default time-to-live is used.
+/// * `truncate_front_to_size` - If the list exceeds this length, remove excess from the front of the list.
+///
+/// # Examples
+/// Assumes that a CacheClient named `cache_client` has been created and is available.
+/// ```
+/// # fn main() -> anyhow::Result<()> {
+/// # use momento_test_util::create_doctest_cache_client;
+/// # tokio_test::block_on(async {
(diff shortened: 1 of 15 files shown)


src/cache/requests/list/list_fetch.rs
```diff
@@ -0,0 +1,225 @@
...
+            .into_inner();
+
+        match response.list {
+            Some(list_fetch_response::List::Missing(_)) => Ok(ListFetch::Miss),
+            Some(list_fetch_response::List::Found(found)) => Ok(ListFetch::Hit {
+                values: ListFetchValue {
+                    raw_item: found.values,
+                },
+            }),
+            _ => unreachable!(),
```
malandis, line 113: In the sorted set implementation we had a case for None, which we interpreted as a `Miss`

src/cache/requests/list/list_concatenate_back.rs
```diff
@@ -0,0 +1,100 @@
+use crate::{
+    cache::{CollectionTtl, MomentoRequest},
+    utils::prep_request_with_timeout,
+    CacheClient, IntoBytes, MomentoResult,
+};
+
+/// Adds multiple elements to the back of the given list. Creates the list if it does not already exist.
```
malandis, line 7: I would add a remark that the order the new values appear in the list is identical to the order of the values passed in.

src/cache/requests/list/list_concatenate_front.rs
```diff
@@ -0,0 +1,100 @@
+use crate::{
+    cache::{CollectionTtl, MomentoRequest},
+    utils::prep_request_with_timeout,
+    CacheClient, IntoBytes, MomentoResult,
+};
+
+/// Adds multiple elements to the front of the given list. Creates the list if it does not already exist.
```
malandis, line 7: Same remark about order applies here.

src/cache/requests/list/list_fetch.rs
```diff
@@ -0,0 +1,225 @@
...
+/// Gets a list item from a cache with optional slices.
+///
+/// # Arguments
+/// * `cache_name` - name of cache
+/// * `list_name` - name of the list
+///
+/// # Optional Arguments
+///
+/// * `start_index` - The starting inclusive element of the list to fetch. Default is 0.
+/// * `end_index` - The ending exclusive element of the list to fetch. Default is end of list.
```
malandis, line 26:
```suggestion
/// * `end_index` - The ending exclusive element of the list to fetch. Default is up to and including end of list.
```

malandis (commented):
Few comments about documentation.

src/cache/requests/list/list_fetch.rs
anitarua: I could return an unknown error instead like so? (and update accordingly for other instances as well)

```
MomentoError {
    message: "Unknown error has occurred, unable to parse list_fetch_response".into(),
    error_code: MomentoErrorCode::UnknownError,
    inner_error: Some(response),
    details: None,
}
```

malandis: Yes that looks good. That follows the JS pattern [here](https://github.com/momentohq/client-sdk-javascript/blob/b0ed1b3ac745ee1eeca1cc8cbc672b1a151268cc/packages/client-sdk-nodejs/src/internal/cache-data-client.ts#L2505-L2515) and [here](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/errors/cache-service-error-mapper.ts#L99). We can extract this to a function later.

anitarua: Made a ticket to do that throughout the sdk: https://github.com/momentohq/client-sdk-rust/issues/239

malandis (approved):

Issue #239, opened 2024-04-30
replace `unreachable` macro with `UnknownError`


anitarua:
Will likely want to add a util function for this.

See original comment thread here: https://github.com/momentohq/client-sdk-rust/pull/237#discussion_r1585237727 

As part of this, audit the responses for other types to handle the case where an `Option` from the server is expected to be `Some` but is `None`.

PR #242, opened 2024-04-30
fix: replace usage of unreachable macro with returning an UnknownError

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/239

Also caught a couple instances where None was being caught but returned a Miss; updated those to return UnknownError as well

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -152,3 +152,15 @@ pub(crate) fn parse_string(raw: Vec<u8>) -> MomentoResult<String> {
         details: None,
     })
 }
+
+pub(crate) fn return_unknown_error(method_name: &str, details: Option<String>) -> MomentoError {
+    MomentoError {
+        message: "Unknown error has occurred, unable to parse ".to_string()
+            + method_name
+            + " : "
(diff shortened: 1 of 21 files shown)


src/utils.rs
```diff
@@ -152,3 +152,15 @@ pub(crate) fn parse_string(raw: Vec<u8>) -> MomentoResult<String> {
...
+    MomentoError {
+        message: "Unknown error has occurred, unable to parse ".to_string()
+            + method_name
+            + " : "
+            + details.as_deref().unwrap_or(""),
+        error_code: MomentoErrorCode::UnknownError,
+        inner_error: None,
+        details: None,
+    }
+}
```
malandis, line 166: Could you make this a function of the `MomentoError` struct? eg

```rust
impl MomentoError {
     pub fn unknown_error(method_name // ..etc) 
```

malandis (commented):
Very nice! Thanks for doing this. Just one suggestion about the location of the utility function.

malandis (approved):

Issue #238, opened 2024-04-30
Root cause rust test failures when number of test threads is high


malandis:
When we leave the number of test threads high, some tests fail. See the integration test target in the makefile for setting the number of test threads. Root cause.

Original comments are here: https://github.com/momentohq/client-sdk-rust/pull/236#issuecommen

# Errors
## Log
Example of test errors when running `cargo t -vv`: [log.txt](https://github.com/momentohq/client-sdk-rust/files/15172586/log.txt). 
## Observation of Errors
No SDK command comes back with an error, rather we actual response type no longer matches the expected. This tests prone to this are conditional set. Instead of `Stored` we get `NotStored` (or vice versa depending on the test)
## Reproducing
When run `cargo t -- --test-threads=2` these errors disappear.

PR #246, opened 2024-05-01
chore: run flush cache test separately from other tests

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/238 

Without limiting number of threads, Michael saw lots of `setIf` test failures, whereas on my machine I saw lots of `list` test failures. I think this is because the `flushCache` test was causing a race condition since it operates on the same test cache as the other tests, but it's nondeterministic when it would kick in.

I used the `#[ignore]` as per the [Rust docs](https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-some-tests-unless-specifically-requested) and added another line to the `test-integration` Makefile target to run the `flushCache` test first, then all the other tests. This allows us to run all the other tests with no thread limitation once again.

I opted to adjust the ordering of tests rather than try to make a new cache for the `flushCache` test to avoid the risk of potentially leaking another cache if the tests fail.

### DIFF
--- a/tests/cache/control.rs
+++ b/tests/cache/control.rs
@@ -57,6 +57,9 @@ mod flush_cache {
         Ok(())
     }
 
+    // This test uses the ignore macro so that we can run it separately from the other
+    // test targets using `cargo test -- --ignored`. This is because all the tests share
+    // the same cache and flushing it will affect the other tests in nondeterministic ways.
     #[tokio::test]
     #[ignore]
     async fn flush_existing_cache_returns_success() -> MomentoResult<()> {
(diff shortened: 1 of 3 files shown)


tests/cache/control.rs
```diff
@@ -58,6 +58,7 @@ mod flush_cache {
     }
 
     #[tokio::test]
+    #[ignore]
```
malandis, line 61: Please leave some documentation here and in the Makefile as to why we do this 👍

malandis (commented):
Quick turnaround! Very nice! Let's leave some documentation for ourselves in the future as to how this works. Documentation around the flush test, in the Makefile, and in the contributing doc would be best.

malandis (approved):

Issue #167, opened 2024-02-15
Add unimplemented set cache methods


nand4011:
The Rust SDK does not support the following set cache methods:

- setAddElement
- setRemoveElement
- setContainsElement
- setContainsElements
- setSample
- setLength
- setPop

anitarua:
setAddElements already exists in cache client, will be implementing setRemoveElements and setFetch to round out the foundational set methods.

PR #248, opened 2024-05-01
feat: add set_fetch and set_remove_elements

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/167

### DIFF
--- a/src/cache/requests/set/set_fetch.rs
+++ b/src/cache/requests/set/set_fetch.rs
@@ -1,0 +1,165 @@
+use std::convert::{TryFrom, TryInto};
+
+use momento_protos::cache_client::set_fetch_response;
+
+use crate::{
+    cache::MomentoRequest,
+    utils::{parse_string, prep_request_with_timeout},
+    CacheClient, IntoBytes, MomentoError, MomentoErrorCode, MomentoResult,
+};
+
+/// Fetch the elements in the given set.
+///
+/// # Arguments
+///
+/// * `cache_name` - The name of the cache containing the set.
+/// * `set_name` - The name of the set to fetch.
+///
+/// # Examples
+/// Assumes that a CacheClient named `cache_client` has been created and is available.
+/// ```
+/// # fn main() -> anyhow::Result<()> {
+/// # use momento::MomentoResult;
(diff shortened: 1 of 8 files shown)


src/cache/requests/set/set_fetch.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+
+impl TryFrom<SetFetchValue> for Vec<Vec<u8>> {
+    type Error = MomentoError;
+
+    fn try_from(value: SetFetchValue) -> Result<Self, Self::Error> {
+        Ok(value.raw_item)
+    }
+}
```
malandis, line 140: This one can be a simple `From` since there is no risk of panic.

src/cache/requests/set/set_fetch.rs
```diff
@@ -0,0 +1,182 @@
...
+            .collect())
+    }
+}
+
+impl TryFrom<SetFetch> for Vec<Vec<u8>> {
+    type Error = MomentoError;
+
+    fn try_from(value: SetFetch) -> Result<Self, Self::Error> {
+        match value {
+            SetFetch::Hit { values } => Ok(values.try_into()?),
```
malandis, line 159: Then this line can just be `Ok(values.into())`

malandis (commented):
Few style nits

src/cache_client.rs
```diff
@@ -556,6 +557,80 @@ impl CacheClient {
...
+    pub async fn set_fetch(
+        &self,
+        cache_name: impl Into<String>,
+        set_name: impl IntoBytes,
+    ) -> MomentoResult<SetFetch> {
+        let request = SetFetchRequest::new(cache_name, set_name);
+        request.send(self).await
+    }
+
+    /// Removes multiple elements from an existing set. If the set is emptied as a result, the set is deleted.
```
malandis, line 596: I also made a note that requesting to remove elements not present is not an error, and requesting to remove elements on a set that doesn't exist is not an error either.

malandis (commented):
One thing to clarify in docs

malandis (approved):

Issue #210, opened 2024-03-27
add dev docs code snippets


anitarua:
To make dev docs site more complete and up to date

PR #253, opened 2024-05-03
chore: add snippets for dev docs, corrected some docstring examples in cache client

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/210

### DIFF
--- a/example/Cargo.toml
+++ b/example/Cargo.toml
@@ -11,3 +11,5 @@ edition = "2021"
 #momento = "0.15.0"
 momento = { path = "../" }
 tokio = { version = "1.18.2", features = ["full"] }
+uuid = { version = "1.2.2", features = ["v4"] }
+futures = "0.3.30"
--- a/example/src/bin/docs_examples.rs
+++ b/example/src/bin/docs_examples.rs
@@ -123,8 +123,8 @@ pub async fn example_API_Increment(
     cache_client: &CacheClient,
     cache_name: &String,
 ) -> Result<(), MomentoError> {
-    cache_client.increment(cache_name, "key", 1).await?;
-    println!("Value incremented");
+    let response = cache_client.increment(cache_name, "key", 1).await?;
+    println!("Value incremented to {}", response.value);
     Ok(())
 }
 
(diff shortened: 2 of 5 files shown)


example/Cargo.toml
```diff
@@ -11,3 +11,5 @@ edition = "2021"
 #momento = "0.15.0"
 momento = { path = "../" }
```
anitarua, line 12: examples are currently using the local SDK, but once we cut a release we can update this here

malandis (commented):
Thanks for doing this. As a discussion point, I see two things:
1. In the examples should we use the ?-operator to propagate errors up from the client calls?
2. Should we test for particular error cases in some examples?

For example in some examples we use the ?-operator on the cache calls, then display the answer in the happy case.

On others we match on the result and display an error.

On others we test for particular error codes, eg:

```rust
    match cache_client.set(cache_name, "key", "value").await {
        Ok(_) => println!("Set successful"),
        Err(e) => {
            if let MomentoErrorCode::NotFoundError = e.error_code {
                println!("Cache not found: {}", cache_name);
            } else {
                eprintln!("Error setting value in cache {}: {}", cache_name, e);
            }
        }
    }
```

My inclination given Rust's features is to:
1. Bubble up the errors using `?`
2. Match on all the other cases (hit vs miss, stored vs not stored, cache created vs already exists)

We can socialize the options with the team during the bug bash.

example/src/bin/docs_examples.rs
```diff
@@ -157,16 +123,8 @@ pub async fn example_API_Increment(
...
-        Err(e) => {
-            if let MomentoErrorCode::NotFoundError = e.error_code {
-                println!("Cache not found: {}", cache_name);
-            } else {
-                eprintln!("Error incrementing value in cache {}: {}", cache_name, e);
-            }
-        }
-    }
+    cache_client.increment(cache_name, "key", 1).await?;
+    println!("Value incremented");
```
malandis, line 127: For this one the response has the new value, which is valuable to print. Can you update the print statement for that?

example/src/bin/docs_examples.rs
```diff
@@ -468,28 +372,50 @@ pub async fn example_API_ListRemoveValue(
...
-pub async fn example_API_DictionarySetFields(
+pub async fn example_API_DictionaryIncrement(
     cache_client: &CacheClient,
     cache_name: &String,
 ) -> Result<(), MomentoError> {
-    match cache_client
+    cache_client
+        .dictionary_increment(cache_name, "dictionary_name", "field", 1)
+        .await?;
+    println!("Incremented field in dictionary");
```
malandis, line 390: Same comment here re: incremented value in response.

malandis (commented):
Couple small issues. Looks great otherwise and much improved and cleaner.

example/src/bin/docs_examples.rs
anitarua: oh good catch, will update

malandis (approved):

PR #262, opened 2024-05-06
chore: port example to new sdk

malandis:
As part of the cutover to the new SDK, we port the example from using the old client to the new one.

### DIFF
--- a/example/Cargo.toml
+++ b/example/Cargo.toml
@@ -8,8 +8,7 @@ edition = "2021"
 [workspace]
 
 [dependencies]
-#momento = "0.15.0"
-momento = { path = "../" }
-tokio = { version = "1.18.2", features = ["full"] }
-uuid = { version = "1.2.2", features = ["v4"] }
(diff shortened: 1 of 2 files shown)


example/Cargo.toml
```diff
@@ -8,8 +8,7 @@ edition = "2021"
 [workspace]
 
 [dependencies]
-#momento = "0.15.0"
-momento = { path = "../" }
-tokio = { version = "1.18.2", features = ["full"] }
-uuid = { version = "1.2.2", features = ["v4"] }
 futures = "0.3.30"
+momento = { path = "../" }
```
anitarua, line 12: should this be `momento = "0.35.0"`?

malandis: Yes it should! Fixed in [6154297](https://github.com/momentohq/client-sdk-rust/pull/262/commits/6154297fb9a442d3464631af0a072f7136a84967)

anitarua (approved):

Issue #259, opened 2024-05-06
verify examples in CI


anitarua:
Examples currently use the local SDK, but after we cut a release with the new methods, we should update the github workflow files to run examples in CI

PR #264, opened 2024-05-06
chore: verify examples in CI and clean up workflow files

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/259


malandis (approved):
Ok with me to remove tests on push to main. Some may feel they are useful if there is a logical error after merging but that's rare. Cc @cprice404

PR #265, opened 2024-05-06
feat: remove legacy sdk and related artifacts

malandis:
With the imminent release of the new SDK, we remove the legacy SDK
from the code base, all related uitlities, tests, and examples.


anitarua (approved):
🧽

src/auth.rs
```diff
@@ -1,216 +0,0 @@
...
-impl From<ChannelConnectError> for AuthError {
-    fn from(value: ChannelConnectError) -> Self {
-        match value {
-            ChannelConnectError::BadUri(err) => Self::BadUri(err),
-            ChannelConnectError::Connection(err) => Self::Connection(err),
-        }
-    }
-}
-
-/// Initiate a login workflow.
```
malandis, line 119: I feel this code belongs in the CLI as opposed to here. The CLI is the sole consumer of this. If we believe that it has broader utility and want to support it with our 1.0 guarantees I can add it back here.

src/cache/requests/scalar/get.rs
```diff
@@ -89,9 +89,9 @@ impl<K: IntoBytes> MomentoRequest for GetRequest<K> {
 ///
 /// If you'd like to handle misses you can simply match and handle your response:
 /// ```
-/// # use momento::response::Get;
+/// # use momento::cache::{Get, GetValue};
 /// # use momento::MomentoResult;
-/// # let get_response = Get::Hit { value: momento::response::GetValue::new(vec![]) };
+/// # let get_response = Get::Hit { value: GetValue::new(vec![]) };
```
malandis, line 94: Many of the docstrings in this file were still referencing the legacy SDK.

Issue #258, opened 2024-05-04
audit documentation and structs to ensure we use key/field/value where appropriate


malandis:
When implementing `dictionary_increment`, I found that the `increment` implementation referred to the cache key to increment as a "field". Audit the types to ensure we use the proper nomenclature.

https://github.com/momentohq/client-sdk-rust/pull/257

PR #266, opened 2024-05-06
chore: audit docs for usage of key/field/value

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/258

### DIFF
--- a/src/cache/cache_client.rs
+++ b/src/cache/cache_client.rs
@@ -1254,7 +1254,7 @@ impl CacheClient {
     /// # let (cache_client, cache_name) = create_doctest_cache_client();
     /// use momento::cache::KeyExists;
     ///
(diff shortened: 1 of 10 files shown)


malandis (commented):
Thanks. Can you also look at the docstrings in the requests?

src/cache/cache_client.rs
```diff
@@ -1275,7 +1275,7 @@ impl CacheClient {
     }
 
     /// Check if the provided keys exist in the cache.
-    /// Returns a list of booleans indicating whether each given key was found in the cache.
+    /// Returns a list or map of booleans indicating whether each given key was found in the cache.
```
malandis, line 1278: It returns one thing which we can convert to either of these right? We should clarify tuat

malandis (approved):

PR #267, opened 2024-05-07
refactor: organize repo

### LINKED ISSUE #261: remove old code from rust sdk
- [ ] Remove old SDK code from Rust SDK
- [ ] Restructure project around new SDK

malandis:
# Overview
This PR refactors the repo organization around the new SDK. Previously the legacy SDK and new SDK comingled. With the legacy SDK out we organize the repo around the new SDK.

If we agree on the following changes, I will document in the contributing file.

To summarize the changes:

# Public API
- We add both the `CacheClient` and `TopicClient` to the library mod.

# Internal API
- We move the `cache_client` and `cache_client_builder` mods under the `cache` mod. The former does not need to be externally accessible; only the `CacheClient` struct needs to be exported. The latter is internal and only used by `CacheClient::builder`.
- We rename `cache::requests` mod to `cache::messages`. Unlike other SDKs, in Rust the requests and responses are more tightly coupled. Because of this the name `requests` does not fully capture the request-response relationship. We rename to `messages` for clarity.
- We move the topic client `Subscription` struct and subscription-related traits and enums under `topics::messages` to parallel the cache organization.

Closes #261


anitarua (approved):
lgtm!

PR #268, opened 2024-05-08
chore: add settings file for rust analyzer env variables

pratik151192:
While testing some of the things the VS Code IDE extensions along with the cargo-lldb debugger has been quite handy. Hooking the env variables for a smoother experience and updated CONTRIBUTING.


anitarua (dismissed):

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "rust-analyzer.runnables.extraEnv": {
+        "MOMENTO_API_KEY": "your_api_key_here"
+    }
+}
```
malandis, line 5: Hmm. I already have this environment variable in my profile, so it's already set. This would clobber that. Also, `settings.json` usually has more IDE/user-specific settings (overrides from user/workspace VS Code settings). Can you rename this file to `.vscode/settings.json.default` or `.template` and the CONTRIBUTING?

malandis (approved):

Issue #271, opened 2024-05-08
`flush_cache` test should create its own cache


anitarua:
Used the `ignore` flag in this fix (https://github.com/momentohq/client-sdk-rust/pull/246) but we actually want to protect the usage of that for future things we actually want to ignore

PR #272, opened 2024-05-08
chore: flush_cache test should create own cache

anitarua:
Addresses #271 

Creating a separate cache for the flush_cache test so that we reserve the `--ignored` flag for test we may actually want to ignore down the road

### DIFF
--- a/tests/cache/control.rs
+++ b/tests/cache/control.rs
@@ -27,6 +27,8 @@ mod create_delete_list_cache {
 }
 
 mod flush_cache {
+    use momento::cache::DeleteCache;
+
     use super::*;
 
     #[tokio::test]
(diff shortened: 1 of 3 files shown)


tests/cache/control.rs
```diff
@@ -57,14 +59,18 @@ mod flush_cache {
         Ok(())
     }
 
-    // This test uses the ignore macro so that we can run it separately from the other
-    // test targets using `cargo test -- --ignored`. This is because all the tests share
-    // the same cache and flushing it will affect the other tests in nondeterministic ways.
+    // Note: the flush_cache test requires creating its own cache as to not interfere with the other integration
+    // tests that all share the same cache. Flushing the cache when other tests are running concurrently creates
+    // a race condition and nondeterministic behavior.
```
malandis, line 64: Need to remove the ignore directive below this

malandis (commented):
One small change to make

malandis (approved):
💥

PR #274, opened 2024-05-08
chore: create examples for topics client

anitarua:
@malandis just putting this PR up to show the current state of the topics client. 

I think we'll probably want to:
- create basic config objects for the TopicClient
- replace `connect` with a `builder` method instead and remove the `user_application_name` parameter
- maybe remove the `resume_at_topic_sequence_number` from the `subscribe` method and create another function that accepts that argument? (sequence number is used only be the resubscribe method, not sure if an end user would ever need to provide that argument)

Thoughts? Anything else that might need to be updated?

### DIFF
--- a/example/Cargo.toml
+++ b/example/Cargo.toml
@@ -9,6 +9,6 @@ edition = "2021"
 
 [dependencies]
 futures = "0.3.30"
-momento = "0.35.0"
+momento = "0.36.0"
 tokio = { version = "1.37.0", features = ["full"] }
 uuid = { version = "1.8.0", features = ["v4"] }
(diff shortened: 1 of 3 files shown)


malandis:
Agreed. We need:
- a builder
- the builder can take the `CredentialProvider`
- the builder also needs to take the topic configuration. For this w need to author the topic client configuration types.
- we can refactor the topic client to have the same `send_request` pattern as the cache client. With that in place we can make the topic sequence number a private field of the struct. We can omit it on `new` and write a setter function. That way users can call `subscribe` without thinking about the sequence number but `resubscribe` can build the request with it and call `send_request`.

Issue #275, opened 2024-05-08
audit topics client for consistency with other SDKs


anitarua:
Audit: https://github.com/momentohq/client-sdk-rust/pull/274
Revision: https://github.com/momentohq/client-sdk-rust/pull/276

Follow-up revisions:

- [x] Rename `SubscriptionRequest` to `SubscribeRequest`
- [x] Update dev docs snippets
- [x] Add topic client example (including subscribing in spawned thread, cleaning up subscription and thread)
- [x] Add more docstrings to all topics-related publicly accessible things 
- [x] Mention importing futures when using topics 
- [x] Mention subscriptions must be `mut` variables in order to call `next()` and consume messages
- [x] Make `SubscriptionItem` publicly accessible
- [x] Double check IDE autocompletes are reasonable
- [x] docstring for subscribe includes old optional sequence number args that aren’t actually there
- [x] docstring for subscribe really needs an example of calling next on the subscription (including the required StreamExt import), it’s very hard to find without that (can be a `no_run` doctest until https://github.com/momentohq/client-sdk-rust/issues/277 is done)

PR #287, opened 2024-05-09
feat: TopicClient follow-up revisions"

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/275

- rename SubscriptionRequest -> SubscribeRequest for consistency
- update readme template to include installing tokio and futures
- make sure all Subscription-related objects are exported
- add more examples in docstrings
- update dev docs examples
- add topics example file

### DIFF
--- a/src/topics/messages/publish.rs
+++ b/src/topics/messages/publish.rs
@@ -22,7 +22,6 @@ use crate::{
 /// # tokio_test::block_on(async {
 /// use momento::{CredentialProvider, TopicClient};
 /// use momento::topics::{TopicPublish, PublishRequest};
-/// use futures::StreamExt;
 ///
 /// let topic_client = TopicClient::builder()
 ///     .configuration(momento::topics::configurations::laptop::latest())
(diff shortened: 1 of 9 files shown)


src/topics/messages/publish.rs
```diff
@@ -17,7 +17,30 @@ use crate::{
 ///
 /// # Example
 ///
-/// See [TopicClient] for an example.
+/// ```
+/// # fn main() -> anyhow::Result<()> {
+/// # tokio_test::block_on(async {
+/// use momento::{CredentialProvider, TopicClient};
+/// use momento::topics::{TopicPublish, PublishRequest};
+/// use futures::StreamExt;
```
malandis, line 25: Is this `use` necessary? (here and throughout)

anitarua: Only necessary when using subscriptions, I can remove for publish-only examples

cprice404 (approved):
yeah

PR #288, opened 2024-05-10
chore: allow `IntoBytesIterable` on statically allocated arrays

malandis:
Broadens the trait implementation for `IntoBytesIterable` from
`Vec<T>` to `IntoIterator<Item = T>` for use on both dynamically
allocated and static arrays.


kvcache (approved):
Yeah, this blanket implementation looks correct to me

Issue #277, opened 2024-05-09
set up integration and docs tests for topic client


anitarua:
Need to set up shared clients for the tests as we did for the cache client

PR #292, opened 2024-05-10
chore: add integration tests and doctests for TopicClient

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/277

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -91,6 +91,15 @@ impl MomentoError {
             details: None,
         }
     }
+
+    pub(crate) fn discontinuity() -> Self {
+        Self {
+            message: "Received a Discontinuity in a Topics subscription".to_string(),
+            error_code: MomentoErrorCode::UnknownError,
+            inner_error: None,
+            details: None,
+        }
+    }
 }
 
 trait ErrorDetails {
--- a/src/topics/messages/subscription.rs
+++ b/src/topics/messages/subscription.rs
@@ -25,16 +25,9 @@ type ChannelType = InterceptedService<Channel, HeaderInterceptor>;
 /// # tokio_test::block_on(async {
 /// use momento::{topics::configurations, CredentialProvider, TopicClient};
 /// use futures::StreamExt;
+/// # let (topic_client, cache_name) = momento_test_util::create_doctest_topic_client();
 ///
-/// let topic_client = TopicClient::builder()
-///     .configuration(configurations::Laptop::latest())
-///     .credential_provider(
-///         CredentialProvider::from_env_var("MOMENTO_API_KEY".to_string())
-///             .expect("API key should be valid"),
-///     )
-///     .build()?;
-///
-/// let mut subscription = topic_client.subscribe("cache", "my-topic").await?;
+/// let mut subscription = topic_client.subscribe(cache_name, "my-topic").await?;
 /// let subscriber_handle = tokio::spawn(async move {
 ///     println!("Subscriber should keep receiving until thread is killed");
 ///     while let Some(message) = subscription.next().await {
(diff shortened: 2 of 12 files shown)


malandis (dismissed):
lgtm

anitarua:
Rebased

malandis (approved):

src/errors.rs
```diff
@@ -91,6 +91,15 @@ impl MomentoError {
             details: None,
         }
     }
```
cprice404, line 93: i know this wasn't part of this PR but why do we have an error for `miss`?

src/topics/messages/subscription.rs
```diff
@@ -286,8 +277,19 @@ pub enum SubscriptionItem {
     Discontinuity(Discontinuity),
```
cprice404, line 277: we don't actually surface the discontinuities to end users in any of the other SDKs yet, do we? I thought that our stance on this was to handle them internally for now. cc: @malandis

test-util/src/test_utils.rs
```diff
@@ -47,10 +48,16 @@ where
...
-    let (cache_client, _) = build_cache_client_and_credential_provider();
+    let (cache_client, _, _) = build_clients_and_credential_provider();
     (cache_client, cache_name)
 }
 
+pub fn create_doctest_topic_client() -> (TopicClient, String) {
+    let cache_name = get_test_cache_name();
+    let (_, topic_client, _) = build_clients_and_credential_provider();
+    (topic_client, cache_name)
+}
```
cprice404, line 59: nice

src/errors.rs
malandis: That's used when you call `try_into` on a response that wasn't a hit.

src/topics/messages/subscription.rs
malandis: @kvcache voiced he wanted to know about discontinuities in his bug bash feedback.

cprice404: @kvcache we have made a deliberate choice to just make this an internal thing in all of the other SDKs so far. We can log them, and we can always expose them more formally as soon as we have a real user with a real use case. I'd like to avoid treating this SDK differently than all of the others at this point though, and we have had no customer use cases that required exposing this yet.

anitarua: I can refactor this then

kvcache: You can’t handle something like a lost message internally. That’s something the user needs a way to handle. Without exposing it in the type model, how is a user supposed to deal with missing messages?

cprice404: there is no way for them to deal with missing messages even if we expose it. This is a 2-way door, we can add it in later if someone needs it; and we may decide that there is a better way to expose it than forcing people to have a branch of their match statement that surfaces this in every place that they are consuming messages.

We don't expose discontinuities at this level in any of the other SDKs and there have not yet been any user inquiries / requests for support for it. Is there a reason we can't wait until we have a real customer use case in hand and then be a little more patient and deliberate in thinking through the options for how to expose it?

kvcache: we need to talk - I believe your opening claim is not true, and adding this branch later is a breaking change XOR you will need to make a weak match semantic on the enum that erodes the compile time safety of the sdk.

cprice404: in that case this whole thing is a bad pattern (stream with match semantic), because i find it hard to believe that we would want to commit to never adding more message types in the future. what if we want to give people a way to access heartbeats at some point? and how confident are you that we would never come up with another type of message we want to surface?

kvcache: This seems like a conversation for meatspace (or video meatspace). It seems fine to address it separately from this pr, but before you label this 1.0, if that helps you make progress more quickly?

cprice404: yeah separate pr is fine

cprice404 (approved):

Issue #279, opened 2024-05-09
Audit usage of `From` vs `TryFrom` implementations


anitarua:
I think there are some response types (e.g. list length, item get ttl, and item get type) that can use `into()` instead of `try_into()` but are missing the `From` implementations

Also Dylan had good feedback:

Hey, I apparently missed this, sorry about that - but from something I was just looking into I noticed there are some places where we're calling .expect in our SDK, e.g. [here](https://github.com/momentohq/client-sdk-rust/blob/647e30dbc5661afea7b3026b02429ec238c76f6f/src/cache/messages/list/list_pop_front.rs#L139)/[here](https://github.com/momentohq/client-sdk-rust/blob/647e30dbc5661afea7b3026b02429ec238c76f6f/src/grpc/header_interceptor.rs#L29)

We should probably be returning an error via the Result type instead, and leaving it up to the user to decide if their application should panic in response

I'd consider adding -W clippy::expect_used to your [lint](https://github.com/momentohq/client-sdk-rust/blob/main/Makefile#L8C1-L8C81) and then using #[allow(clippy::expect_used)] to exempt usages if you have a legitimate reason

PR #293, opened 2024-05-10
chore: add lint rule to check usage of expect

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/279

Added the clippy rule Dylan recommended and refactored as needed. Had to revise the `lint` Makefile and add `#[allow(clippy::expect_used)]` in several places to omit test-related files from the rule as we _do_ want the tests to panic if something goes wrong.

Current implementations of `From` and `TryFrom` are fine upon inspection.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,12 @@
 all: precommit
 
 .PHONY: lint
+## Check the formatting of all files, run clippy on the source code, then run
+## clippy on the tests (but allow expect to be used in tests)
 lint:
 	cargo fmt -- --check
-	cargo clippy --all-targets --all-features -- -D warnings -W clippy::unwrap_used
+	cargo clippy --all-features -- -D warnings -W clippy::unwrap_used -W clippy::expect_used
+	cargo clippy --tests -- -D warnings -W clippy::unwrap_used
 
 .PHONY: build
 ## Build project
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -3,9 +3,12 @@
 all: precommit
 
 .PHONY: lint
+## Check the formatting of all files, run clippy on the source code, then run
+## clippy on the tests (but allow expect to be used in tests)
 lint:
 	cargo fmt -- --check
-	cargo clippy --all-targets --all-features -- -D warnings -W clippy::unwrap_used
+	cargo clippy --all-features -- -D warnings -W clippy::unwrap_used -W clippy::expect_used
+	cargo clippy --tests -- -D warnings -W clippy::unwrap_used
```
cprice404, line 11: if you are doing `allow` in the test code where we are using `expect`, then you don't need to split these here, do you?

cprice404 (commented):
looks good except the one question about the makefile

Makefile
anitarua: I used the `allow` in the test setup and teardown files since those are marked as "not tests" [here](https://github.com/momentohq/client-sdk-rust/blob/main/Cargo.toml#L18-L30), but I didn't want to go in and add the `allow` everywhere we used `expect` in the test files

anitarua:
rebased

Makefile
cprice404: :ack:

cprice404 (dismissed):

anitarua:
rebased

cprice404 (approved):
expect

Issue #284, opened 2024-05-09
Improve docs


anitarua:
- [x] Add [lint rule](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#missing-docs) for missing docs -> `-W missing_docs` -> do this last to narrow down the warnings
- [x] Add `//!` docs to top-level modules (other modules too if needed?)
- [x] Some dictionary docs in cache_client are missing links to their *Request counterparts
- [x] List length doc example: `cache_clientlist_length` -> `cache_client.list_length` (didn't see it)
- [x] Request docs examples should exercise optional arguments (list concat front/back, probably others)
- [x] `send_request` should link to sorted set fetch by score example for optional fields instead because we changed fetch by rank to accept Optional args in the `cache_client` function signature
- [x] Note that `tokio` works best with grpc in readme, our recommended async runtime, and should be installed when installing the Momento SDK
- [x] For return types that support Hit/Miss, we should include an example in the docstring showing `try_into` directly on the response object, without a `match`, to convert to `HashMap` or whatever, in addition to showing the `match` approach.

PR #294, opened 2024-05-10
feat: add list_push_front and list_push_back

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/278

Also fixes a couple of the list-related tasks in https://github.com/momentohq/client-sdk-rust/issues/284


cprice404 (approved):
chain

PR #304, opened 2024-05-10
chore: add top-level docs for the crate, including 'into' conventions etc.

cprice404:


src/lib.rs
```diff
@@ -1,3 +1,86 @@
...
+//! into your desired type.
+//! 
+//! All Momento cache values are stored as `vec<u8>`, but if you are using UTF-8 strings, you can use `try_into`
+//! for these coercions as well.
+//! 
+//! Here are a few examples of how you can interact with a `CacheClient::get` response:
+//! 
+//! Using a `match`:
+//! 
+//! ```
```
anitarua, line 59: I think you need to specify `no_run` on these code blocks

src/errors.rs
```diff
@@ -102,31 +102,6 @@ impl MomentoError {
...
-    }
-
-    fn inner_error(&self) -> Option<&ErrorSource> {
-        self.inner_error.as_ref()
-    }
-
-    fn details(&self) -> Option<&MomentoGrpcErrorDetails> {
-        self.details.as_ref()
-    }
-}
```
cprice404, line 128: @anitarua I didn't especially want to delete this stuff, but I was getting a compiler warning before, and then when I started trying to run `make` it was giving me a full-on error about this. not sure what this stuff is for or if it's really okay to delete it.

anitarua: I think I had basically made a duplicate called `MomentoGrpcErrorDetails` when I worked on the error stuff a while ago, I think it should be find to delete this

cprice404: okay. idk why this is failing for me but not for y'all :)

src/lib.rs
anitarua: looks like you need to set up the whole doctest client if you want it to compile, or just `ignore` to make the doctests skip over it ([reference](https://stackoverflow.com/questions/32429369/how-to-opt-out-of-running-a-doc-test))

cprice404: yeah working on adding the client stuff, ty

src/lib.rs
```diff
@@ -1,3 +1,118 @@
+//!
+//! # Momento
+//! Welcome to the Momento Rust SDK!
+//!
+//! This crate contains a `CacheClient` for interacting with a serverless Momento Cache, and a
+//! `TopicClient` for interacting with serverless Momento Topics (pub/sub messaging). For more detailed
+//! information see the [Momento documentation](https://docs.momentohq.com).
+//!
+//! Construct instances of these clients using `CacheClient::builder()` and `TopicClient::builder()`.
```
anitarua, line 9: Might be nice to turn these into links to the builders

```suggestion
//! Construct instances of these clients using [CacheClient::builder()] and [TopicClient::builder()].
```

src/lib.rs
```diff
@@ -1,3 +1,118 @@
...
+//!
+//! - `momento::cache::configurations::Laptop::latest()` - suitable for a development environment with lenient timeouts
+//! - `momento::cache::configurations::InRegion::latest()` - suitable for a production configuration with more strict timeouts.
+//!
+//! These configurations can be passed to the `CacheClient` and `TopicClient` builders.
+//! For advanced use cases you can build your own configurations rather than using the pre-builts.
+//!
+//! ## Credential Providers
+//!
+//! The `CredentialProvider` struct is used to provide the API key for the Momento service. The two
```
anitarua, line 34:
```suggestion
//! The [CredentialProvider] struct is used to provide the API key for the Momento service. The two
```

src/lib.rs
cprice404: Thank you! I didn't know that syntax.

anitarua (approved):
🎉

malandis (approved):
❤️

PR #315, opened 2024-05-13
refactor: rename various response value types to `Value` and update traits

### LINKED ISSUE #285: Revise model for response types?
After discussing, we're going to go ahead and rename response types to include `Response` at the end.

----------

cc: @cprice404 @malandis 

Chris: for response types that only have a single non-error case, like TopicPublish and DictionarySetFields, I can’t decide how I feel about our current representation. I like that we have a struct for these cases, in case we need to add fields in the future like fieldCount or any other metadata. But right now it feels like I can’t access that struct without a match, and that match feels confusing and superfluous. I suspect that the argument in favor of this is to make it symmetrical with the response types that have more than one success case, like Hit/ Miss, but after typing some code against this SDK it felt so weird to have to put that match there that I kind of don’t think the symmetry is a net win. I think this is a place where, in languages like Rust/Golang that have a formal built-in pattern for returning errors along with a return value, we will likely just want to shed the outer type for a Success response and avoid forcing people to do an extra match. This might warrant a quick discussion with some interested parties to get aligned if there is not an obvious consensus.

Kenny:
this completion is hard to guess which one is right
Untitled

Can rename them to `*Response` ? Or might be helped by removing structs for non-error cases

malandis:
Renames various collection inner value types to `Value` and updates conversion traits.

None of these types (except one) need to be rolled up to `momento::cache`. They are either only used in:
- doctests, or
- integration tests

In both of those cases we can avoid using them by using various `From` and `Default` traits. We keep the types publicly visible so a user can still `use` them, albeit with a longer fully qualified name.

The only inner value type we roll up to `momento::cache` is `SortedSetElements`. We still roll this one up since as user may want to pass a `Vec` of these to a `SortedSetPutElementsRequest`.

Closes #285

### DIFF
--- a/src/cache/messages/data/dictionary/dictionary_fetch.rs
+++ b/src/cache/messages/data/dictionary/dictionary_fetch.rs
@@ -87,7 +87,7 @@ impl<D: IntoBytes> MomentoRequest for DictionaryFetchRequest<D> {
                     .map(|element| (element.field, element.value))
                     .collect();
                 Ok(DictionaryFetchResponse::Hit {
-                    value: DictionaryFetchValue::new(raw_item),
+                    value: Value::new(raw_item),
                 })
             }
             _ => Err(MomentoError::unknown_error(
@@ -104,9 +104,9 @@ impl<D: IntoBytes> MomentoRequest for DictionaryFetchRequest<D> {
 /// ```
 /// fn main() -> anyhow::Result<()> {
 /// # use std::collections::HashMap;
-/// # use momento::cache::{DictionaryFetchResponse, DictionaryFetchValue};
+/// # use momento::cache::DictionaryFetchResponse;
 /// # use momento::MomentoResult;
-/// # let fetch_response = DictionaryFetchResponse::Hit { value: DictionaryFetchValue::default() };
+/// # let fetch_response = DictionaryFetchResponse::default();
 /// use std::convert::TryInto;
 /// let item: HashMap<String, String> = match fetch_response {
 ///    DictionaryFetchResponse::Hit { value } => value.try_into().expect("I stored strings!"),
@@ -119,9 +119,9 @@ impl<D: IntoBytes> MomentoRequest for DictionaryFetchRequest<D> {
 /// Or, if you're storing raw bytes you can get at them simply:
 /// ```
 /// # use std::collections::HashMap;
-/// # use momento::cache::{DictionaryFetchResponse, DictionaryFetchValue};
+/// # use momento::cache::DictionaryFetchResponse;
 /// # use momento::MomentoResult;
-/// # let fetch_response = DictionaryFetchResponse::Hit { value: DictionaryFetchValue::default() };
+/// # let fetch_response = DictionaryFetchResponse::default();
 /// use std::convert::TryInto;
 /// let item: HashMap<Vec<u8>, Vec<u8>> = match fetch_response {
 ///   DictionaryFetchResponse::Hit { value } => value.into(),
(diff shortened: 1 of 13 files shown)


anitarua (dismissed):
🎉

src/cache/messages/data/dictionary/dictionary_fetch.rs
```diff
@@ -136,43 +136,52 @@ impl<D: IntoBytes> MomentoRequest for DictionaryFetchRequest<D> {
...
 /// let item: MomentoResult<HashMap<Vec<u8>, Vec<u8>>> = fetch_response.try_into();
 /// ```
 #[derive(Debug, PartialEq, Eq)]
 pub enum DictionaryFetchResponse {
-    Hit { value: DictionaryFetchValue },
+    Hit { value: Value },
     Miss,
 }
 
+impl Default for DictionaryFetchResponse {
```
cprice404, line 161: does it make sense to have a default for this?

cprice404 (approved):
code changes lgtm if they look good to @anitarua .

I'm not really sure about the `default`s - are those publicly visible? They seem like they could be useful for testing / internal stuff, but weird for an end user to be able to see.

src/cache/messages/data/dictionary/dictionary_fetch.rs
malandis: I implemented `Default` for various response types to make the doctests easier to write. In hindsight this will be confusing on the public API. I can back out those changes.

src/cache/messages/data/set/set_fetch.rs
```diff
@@ -107,41 +105,47 @@
...
 }
 
-impl TryFrom<SetFetchValue> for Vec<String> {
+impl TryFrom<Value> for Vec<String> {
     type Error = MomentoError;
 
-    fn try_from(value: SetFetchValue) -> Result<Self, Self::Error> {
-        value.raw_item.into_iter().map(parse_string).collect()
+    fn try_from(value: Value) -> Result<Self, Self::Error> {
+        Ok(value.raw_item.into_iter().map(parse_string).collect())
```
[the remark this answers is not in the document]
malandis: Fixed in [a5a4801](https://github.com/momentohq/client-sdk-rust/pull/315/commits/a5a4801bbe45a2a41830867af1624b314e76d5d7)

src/cache/messages/data/list/list_fetch.rs
```diff
@@ -141,43 +139,49 @@
...
 }
 
-impl TryFrom<ListFetchValue> for Vec<String> {
+impl TryFrom<Value> for Vec<String> {
     type Error = MomentoError;
 
-    fn try_from(value: ListFetchValue) -> Result<Self, Self::Error> {
-        value.raw_item.into_iter().map(parse_string).collect()
+    fn try_from(value: Value) -> Result<Self, Self::Error> {
+        Ok(value.raw_item.into_iter().map(parse_string).collect())
```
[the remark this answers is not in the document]
malandis: Fixed in [a5a4801](https://github.com/momentohq/client-sdk-rust/pull/315/commits/a5a4801bbe45a2a41830867af1624b314e76d5d7)

anitarua (dismissed):

src/cache/messages/data/dictionary/dictionary_fetch.rs
malandis: Fixed in [4941772](https://github.com/momentohq/client-sdk-rust/pull/315/commits/494177262484f075c5454cf6db2d13e6b1f9ec91)

cprice404 (approved):
value

Issue #299, opened 2024-05-10
what to do about discontinuities


cprice404:

anitarua:
We discussed with Kenny and determined we'll eventually have two subscribe methods:
* The basic `subscribe` method will return only values (string or binary), meaning we can remove the `Discontinuity` type and a pattern matching layer from the Rust SDK for now
* The advanced method (to be named) that will return `SubscriptionItem`s that include values, discontinuities, heartbeats, etc future messages we want to return

PR #316, opened 2024-05-13
feat: basic subscriptions should return only SubscriptionValues

anitarua:
Addresses https://github.com/momentohq/client-sdk-rust/issues/299

- The basic `subscribe` method returns only `SubscriptionValue`s now (only binary or string values, not discontinuities)
- Removed the discontinuity error and `TryFrom` that had been used in a test
- Renamed `TopicPublish` to `TopicPublishResponse` for consistency with the rest of the SDK
- Updated doctests and filled in more missing docstrings related to topics

### DIFF
--- a/src/topics/messages/subscription.rs
+++ b/src/topics/messages/subscription.rs
@@ -1,5 +1,5 @@
 use core::str;
-use std::convert::{TryFrom, TryInto};
+use std::convert::TryFrom;
 
 use futures::future::BoxFuture;
 use futures::{Future, FutureExt};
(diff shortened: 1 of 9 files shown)


src/topics/messages/subscription.rs
```diff
@@ -36,7 +36,10 @@ type ChannelType = InterceptedService<Channel, HeaderInterceptor>;
 /// let subscriber_handle = tokio::spawn(async move {
 ///     println!("Subscriber should keep receiving until thread is killed");
 ///     while let Some(message) = subscription.next().await {
-///         println!("[1] Received message: {:?}", message);
+///         match message.kind {
+///             momento::topics::ValueKind::Text(t) => println!("Received message as string: {:?}", t),
+///             momento::topics::ValueKind::Binary(b) => println!("Received message as bytes: {:?}", b),
+///         }
```
anitarua, line 42: updated doctest to show how to pattern match on the `SubscriptionValue`

cprice404 (approved):
simple

PR #319, opened 2024-05-13
chore: move docs examples into subdir

cprice404:
This commit moves the docs examples into a different subdir rather
than being in the `bin` dir. My concern is that it might be confusing
for users who were looking around in the bin dir and stumbled across
those docs examples. I had previously tried putting them into a library
subdir, but that was causing cargo/clippy to fail due to dead code
and other issues.

This commit updates the cargo.toml file to treat them as `bin` executables,
which gives them a different scope for clippy and allows them to pass
the compile/lint phases without their `main` being flagged as "dead code".

Will update the dev docs accordingly after this is merged.


bruuuuuuuce (approved):
:shipit:

malandis (approved):

Makefile
```diff
@@ -59,7 +59,7 @@ build-examples:
 .PHONY: run-examples
 ## Run example code
 run-examples:
-	cd example && make lint && cargo run --bin=cache && cargo run --bin=readme && cargo run --bin=docs_examples && cargo run --bin=topics
+	cd example && make lint && cargo run --bin=readme && cargo run --bin=cache && cargo run --bin=topics && cargo run --bin=docs_examples && cargo run --bin=cheat_sheet_client_instantiaion
```
malandis, line 62:
```suggestion
	cd example && make lint && cargo run --bin=readme && cargo run --bin=cache && cargo run --bin=topics && cargo run --bin=docs_examples && cargo run --bin=cheat_sheet_client_instantiation
```

malandis (approved):

PR #323, opened 2024-05-14
feat: implement custom debug and display traits for byte-array types

### LINKED ISSUE #282: Audit debug printing
Bug bashers noted that they expected `GetValue` to print strings, but instead saw `raw_item: [ bytes ]`

We should probably refer to the dotnet SDK and add util methods to truncate really long raw values and show what they'd look like when we convert bytes to strings

Should also audit usage of the `Debug` macro and make sure it's available on everything that users might want to print out

malandis:
Presents a pattern we can use to get debug and display for response types.

Because our cache stores data as bytes, when users build a request or receive a response, they get back a wrapper around `Vec<u8>` of some form. This may be a `Vec<u8>` directly or another data structure like a `HashMap` or another `Vec`, for example.

In order to aid debugging and printing, we provide debug and display trait implementations that attempt to interpret and display the bytes as strings. There are two aspects to this:

1. For each `Value` type in a response, we manually implement `std::fmt::Debug` and `std::fmt::Display`. We have authored utility functions in `utils::fmt` to aid with this. See the `get_response::Value` implementation.
2. We auto-derive `Debug` and `Display` for the response type (`GetResponse` in this PR). Unfortunately Rust does not allow one to auto-derive `Display`, so we had to use the third-party library `derive_more` for that. Hence this PR adds that as a dependency.

Going forward, we can implement debug and display as appropriate for the `Value` types and auto-derive `Debug` and `Display` for the response types. We can also add truncation in the utility functions.

Closes #282

### DIFF
--- a/src/cache/messages/data/scalar/get.rs
+++ b/src/cache/messages/data/scalar/get.rs
@@ -1,6 +1,5 @@
 use crate::cache::messages::MomentoRequest;
-use crate::utils::parse_string;
-use crate::utils::prep_request_with_timeout;
+use crate::utils;
 use crate::CacheClient;
 use crate::{IntoBytes, MomentoError, MomentoResult};
 use momento_protos::cache_client::ECacheResult;
@@ -57,7 +56,7 @@ impl<K: IntoBytes> MomentoRequest for GetRequest<K> {
     type Response = GetResponse;
 
     async fn send(self, cache_client: &CacheClient) -> MomentoResult<GetResponse> {
-        let request = prep_request_with_timeout(
+        let request = utils::prep_request_with_timeout(
             &self.cache_name,
             cache_client.configuration.deadline_millis(),
             momento_protos::cache_client::GetRequest {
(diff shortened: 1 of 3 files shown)


src/cache/messages/data/scalar/get.rs
```diff
@@ -152,30 +151,36 @@ impl<I: IntoBytes> From<I> for GetResponse {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        utils::write_bytes_for_debug(f, "raw_item", &self.raw_item)
+    }
+}
+
+impl std::fmt::Display for Value {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
```
malandis, line 170: With this implementation, when I run the example and print the `GetResponse`, I get:

```bash
Listing caches:
my-test-cache
cache

Setting key: my_key, value: my_value
Got value: Ok(Hit { value: raw_item: [109, 121, 95, 118, 97, 108, 117, 101] (as string: "my_value") })
Got value: my_value
Permanently deleted cache named, cache
```

malandis (commented):
Thread for initial review comments

src/cache/messages/data/scalar/get.rs
anitarua: I think that looks good! And the `<invalid UTF-8>` error case seems reasonable to me

malandis:
> looks like multiline strings have a syntax in Rust: https://rustjobs.dev/blog/formatting-multiline-strings-with-interpolation-in-rust/

As expected, I never would have guessed that was the syntax. 😆

anitarua (approved):
🎉

cprice404 (approved):

cprice404:
if you already have this implemented for a dictionary or other collection type, feel free to include the example for that! but if not no worries

malandis:
> if you already have this implemented for a dictionary or other collection type, feel free to include the example for that! but if not no worries

I have not, but it should be straightforward to model after the `utils::write_bytes_for_debug`

PR #332, opened 2024-05-17
chore: run ci on ubuntu-24.04

malandis:
Previously we ran on `macos` to sidestep timeouts of uncertain origin on `ubuntu-22.04` + GitHub's Azure runners. This has resolved on `ubuntu-24.04`


rishtigupta (approved):
:woot

PR #335, opened 2024-05-20
chore: implement Debug for DictionaryFetchResponse, tweak debug fmt utils

cprice404:
I wanted to play with what it would feel like to implement Debug for
other response types. In this PR I implemented it for DictionaryFetchResponse.

However, along the way I found that it was really hard to get the pretty-print
indenting correct when we are manually calling `Formatter::write*`; it's much
easier if we can stick to calling the `Formatter::debug_*` variants. So
I refactored the `fmt` utils a bit, to give us a `DebuggableValue` trait. This
trait provides an `as_debuggable_value` function for vec<u8>, String, and &str.
The implementations of Debug for the response Value structs can just use this
to get an instance of DebuggableValue, and then use `Formatter.debug_*` to
debug whatever collection type is appropriate.

### DIFF
--- a/src/cache/messages/data/scalar/get.rs
+++ b/src/cache/messages/data/scalar/get.rs
@@ -1,5 +1,6 @@
 use crate::cache::messages::MomentoRequest;
 use crate::utils;
+use crate::utils::fmt::AsDebuggableValue;
 use crate::CacheClient;
 use crate::{IntoBytes, MomentoError, MomentoResult};
 use derive_more::Display;
@@ -160,9 +161,11 @@ pub struct Value {
 
 impl std::fmt::Debug for Value {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        utils::fmt::write_struct_begin(f, "Value")?;
-        utils::fmt::write_bytes_for_debug(f, "raw_item", &self.raw_item)?;
-        utils::fmt::write_struct_end(f)
+        let raw_item = &self.raw_item;
+        let debug_value = raw_item.as_debuggable_value();
+        f.debug_struct("Value")
+            .field("raw_item", &debug_value)
+            .finish()
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/cache/messages/data/scalar/get.rs
```diff
@@ -160,9 +161,11 @@ pub struct Value {
...
 impl std::fmt::Debug for Value {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        utils::fmt::write_struct_begin(f, "Value")?;
-        utils::fmt::write_bytes_for_debug(f, "raw_item", &self.raw_item)?;
-        utils::fmt::write_struct_end(f)
+        let raw_item = &self.raw_item;
+        let debug_value = raw_item.as_debuggable_value();
+        f.debug_struct("Value")
+            .field("raw_item", &debug_value)
+            .finish()
```
malandis, line 168: That's way nicer!

src/cache/messages/data/scalar/get.rs
```diff
@@ -224,22 +227,49 @@ mod tests {
         let hit = GetResponse::Hit {
             value: Value::new("hello".as_bytes().to_vec()),
         };
+        assert_eq!(format!("{}", hit), r#"Value { raw_item: "hello" }"#);
         assert_eq!(
             format!("{:?}", hit),
-            "Hit { value: Value { raw_item: [104, 101, 108, 108, 111] (as string: \"hello\") } }"
+            r#"Hit { value: Value { raw_item: "hello" } }"#
```
malandis, line 233: My only concern here is there could be binary values that can render as utf-8 by happenstance. That said I don't think the raw bytes are all that informative anyway so 🤷

src/utils.rs
```diff
@@ -174,50 +174,58 @@ where
...
+                Err(_) => DebuggableValue::Bytes(self.clone()),
+            }
+        }
+    }
+
+    impl Debug for DebuggableValue {
+        fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+            match self {
+                DebuggableValue::String(s) => f.write_fmt(format_args!("{:?}", s)),
+                DebuggableValue::Bytes(b) => f.debug_list().entries(b.iter()).finish(),
```
malandis, line 227: Nice

malandis (approved):
Those formatter functions are slick!

anitarua (approved):

PR #338, opened 2024-05-22
chore: move sdk source code to a subdir

cprice404:
Having Cargo.toml at the root causes problems where other
Cargo.toml files (e.g. in the examples dir) end up thinking that
they are part of a cargo workspace. This is causing problems
when trying to build lambda examples.

It's better for the SDK's Cargo.toml to not exist in a parent dir
of any Cargo.toml files that are not related to the SDK library
itself. To that end, this commit moves all of the SDK code into
a subdir called `sdk`. The `Makefile` is still at the root, and
the targets have been updated to descend into the sdk directory
before executing.

Also removed the Cargo.lock files from .gitignore.


malandis (approved):
Nice to centralize all the changes to the `Makefile` 👍

PR #355, opened 2024-06-04
fix: Move tests directory into sdk

nand4011:
Move the tests directory into the sdk directory so the project can run them.


cprice404 (approved):
:ty:

do we need to do anything to the makefile?

cprice404:
> :ty:
> 
> do we need to do anything to the makefile?

looks like no

🚢

nand4011:
We don't. The makefile commands all run from the sdk directory, and the project assumes there will be a tests directory at the same level as Cargo.toml, so it works.

nand4011:
It does look like the unit tests and integration tests can't be distinguished, so the unit tests step is redundant. It might be worth making a ticket to separate them more.

PR #357, opened 2024-06-06
chore: make optional things optional on request builders

### LINKED ISSUE #356: allow options on builders
For ease of use, make request builders accept `impl Into<Option<>>` where appropriate.

malandis:
Per user feedback, we should support `Option` parameters on builders when the parameter is actually optional. Currently the SDK expects mandatory values for those builder functions, even when the parameter is actually optional.

To do this, where applicable, we change each parameter type to be `impl Into<Option<>>` which is backwards compatible.

Closes #356

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,11 @@
 ## Generate sync unit tests, format, lint, and test
 all: precommit
 
+.PHONY: format
+## Format all files
+format:
+	cd sdk && cargo fmt
+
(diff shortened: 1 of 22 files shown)


nand4011 (approved):
LGTM.
Is this to make it more visually clear that these are optional arguments? I don't know why you would call the builder function for an optional argument if you were supplying None.

PR #365, opened 2024-06-20
feat: improve NotFound handling, return Optional for storage get

cprice404:
This is a POC for addressing some of the things that we ran into in the other SDKs; namely:

1. Split up NotFound errors into distinct categories
2. Change GetResponse value to Optional, to allow us to distinguish between NotFound and WrongType

Also removes the inner `Success` enum value from the GetResponse because it was making it harder to work with the success case, and in the Cache client, we don't use enums when we only have a single response type.

Open to feedback on all of the above.

### DIFF
--- a/example/rust/Cargo.toml
+++ b/example/rust/Cargo.toml
@@ -17,7 +17,9 @@ path = "src/docs_examples/docs_examples.rs"
 
 [dependencies]
 futures = "0.3.30"
-momento = { version = "0.39.0" }
+momento = { version = "0.39.0", path = "../../sdk" }
 tokio = { version = "1.37.0", features = ["full"] }
 uuid = { version = "1.8.0", features = ["v4"] }
 anyhow = "1"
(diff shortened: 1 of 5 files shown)


anitarua (approved):
lgtm!

example/rust/Cargo.toml
```diff
@@ -17,7 +17,9 @@ path = "src/docs_examples/docs_examples.rs"
 
 [dependencies]
 futures = "0.3.30"
-momento = { version = "0.39.0" }
+momento = { version = "0.39.0", path = "../../sdk" }
```
nand4011, line 20: I assume this and the example are there for development and shouldn't be merged.

nand4011: I'll merge as is and revert in the other pr.

cprice404: correct, i just wanted you to be able to see what the example code looks like and make sure you think it's sane

PR #368, opened 2024-06-21
feat: add example for preview storage client

pratik151192:
I may have to leave the bug bash a little early so started early with the Rust SDK. This might have been in your backlog/WIP already but since I was about to follow examples pattern anyway, just decided to PR it.

Note that I did NOT update the CI workflows to run this due to alpha instability.

```
my_momento_store was created
Key foo was successfully persisted in my_momento_store
Key foo found in my_momento_store with value String("bar")
Key foo was successfully deleted from my_momento_store
Store was deleted
```

### DIFF
--- a/example/rust/src/bin/storage.rs
+++ b/example/rust/src/bin/storage.rs
@@ -24,11 +24,10 @@ async fn main() -> Result<(), MomentoError> {
 
     // List all stores and validate my_momento_store was created
     let list_stores_response = storage_client.list_stores().await?;
-    if list_stores_response
+    if !list_stores_response
         .stores
         .iter()
         .any(|x| x.name == store_name)
-        == false
     {
         eprintln!("{store_name} was not created");
         process::exit(1);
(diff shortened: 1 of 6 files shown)


example/rust/src/bin/storage.rs
```diff
@@ -0,0 +1,109 @@
...
+            } else {
+                eprintln!("Key {key} was not found!");
+                process::exit(1);
+            }
+        }
+        Err(err) => {
+            eprint!("error while getting key: {err}");
+            process::exit(1);
+        }
+    }
```
malandis, line 65: You may want to show pattern matching over the type to demonstrate how to handle a value of unknown type.

example/rust/src/bin/storage.rs
```diff
@@ -0,0 +1,109 @@
...
+    // Validate store was deleted
+    let list_stores_response = storage_client.list_stores().await?;
+    if list_stores_response
+        .stores
+        .iter()
+        .any(|x| x.name == store_name)
+    {
+        eprintln!("{store_name} was not deleted");
+        process::exit(1);
+    }
```
malandis, line 104: This is duplicated from above. Let's add a helper function and extract it.

malandis (commented):
Looks good just a couple minor comments

example/rust/src/bin/storage.rs
pratik151192: Discussed this on Slack. need an export for `StoreValue`

example/rust/src/bin/storage.rs
pratik151192: Done

cprice404 (approved):

cprice404:
pushed a few commits to rebase and add to CI

malandis (approved):

PR #370, opened 2024-06-25
fix: export StorageValue and align the exports with the cache client.

nand4011:
Rename StoreValue to StorageValue.

Export StorageValue so it can be matched on.

Bring the storage client exports in line with the cache client exports.

Match on item_not_found instead of element_not_found.

Add doc strings for the now public modules and files.

### DIFF
--- a/sdk/src/storage/messages/mod.rs
+++ b/sdk/src/storage/messages/mod.rs
@@ -1,7 +1,7 @@
-/// Control messages for storage
+/// Control plane messages for storage
 pub mod control;
 
-/// Data messages for storage
+/// Data plane messages for storage
 pub mod data;
 mod momento_storage_request;
 pub use momento_storage_request::MomentoStorageRequest;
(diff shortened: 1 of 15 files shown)


sdk/src/storage/messages/mod.rs
```diff
@@ -1,5 +1,9 @@
+/// Control messages for storage
 pub mod control;
 
+/// Data messages for storage
```
bruuuuuuuce, line 4:
```suggestion
/// Data plane messages for storage
```

sdk/src/storage/messages/mod.rs
```diff
@@ -1,5 +1,9 @@
+/// Control messages for storage
```
bruuuuuuuce, line 1:
```suggestion
/// Control plane messages for storage
```

bruuuuuuuce (approved):
:shipit:

PR #376, opened 2024-07-03
fix: storage get response should use Found and NotFound cases

anitarua:


malandis (commented):
Does not found error correspond to item not found (vs store)?

anitarua:
Yeah, not found error indicates item not found

malandis (approved):

PR #383, opened 2024-07-08
chore: add vscode workspace

malandis:
To make development on the sdk and various examples easier, we add a
workspace file. Other developers can add more example projects as necessary.


cprice404 (approved):
historically i've always been told / believed that IDE config files should not be checked in to VCS at all, but if this is common for vscode, 🤷

malandis:
> historically i've always been told / believed that IDE config files should not be checked in to VCS at all, but if this is common for vscode, 🤷

This is an opt-in feature from VS Code. Instead of loading a particular folder, you can load a "workspace", which has a higher level grouping of the projects within.

PR #386, opened 2024-07-12
chore: update examples momento dep and update storage example get responses

anitarua:
An addition to https://github.com/momentohq/dev-eco-issue-tracker/issues/936

Updated the storage get responses but also needed to update them in the storage example


cprice404 (approved):
dance

PR #391, opened 2024-07-22
refactor: lazily initialize control client

malandis:
# Overview

Demonstrates how to lazily initialize the control client. Stores the
control client as an `Arc<Mutex<Option<ScsControlClient<>>>>`:

- Since the cache client is cloneable, Arc allows it to be cloned
- Mutex for exclusive access to lazily initialize only once
- Option to track the state from uninitialized to initialized

A helper function `get_control_client` does
get-or-lazily-initialize. Uses another helper `create_control_client`
to instantiate it.

In this PR we've refactored the control requests to use this. The data
client refactor would be nearly identical but would take changes to
many request files. So we exclude that from this spike.

We could refactor this into its own component for re-use. It would parallel
the `once_cell::sync::Lazy` struct (see below).

# Alternatives considered

There is a rust library called `once_cell` which has a `Lazy` type that takes an
initializer function. This has to be a pure function and cannot capture variables
from the outer scope (which we would need for the credential provider), so
we could not get that to work.

### DIFF
--- a/sdk/src/cache/cache_client.rs
+++ b/sdk/src/cache/cache_client.rs
@@ -1,4 +1,5 @@
 use std::convert::TryInto;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use momento_protos::cache_client::scs_client::ScsClient;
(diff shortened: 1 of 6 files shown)


sdk/src/cache/cache_client.rs
```diff
@@ -118,6 +123,47 @@ impl CacheClient {
...
+            ),
+        );
+
+        Ok(ScsControlClient::new(control_interceptor))
+    }
+
+    pub(crate) fn get_control_client(
+        &self,
+    ) -> MomentoResult<ScsControlClient<InterceptedService<Channel, HeaderInterceptor>>> {
+        let mut control_client = match self.control_client.lock() {
```
cprice404, line 151: imo the relatively minor API annoyance of having the builder return a `Result` is justified if we would have to introduce mutexes and locks in order to avoid it. wdyt

malandis: Yes I'd rather have the clients available upon building the clients. This is a lot of work for lazy initialization.

PR #393, opened 2024-10-08
feat: adding support for multiple connections

cprice404:


sdk/src/cache/cache_client.rs
```diff
@@ -76,12 +77,14 @@ use crate::{utils, IntoBytes, MomentoResult};
...
 #[derive(Clone, Debug)]
 pub struct CacheClient {
-    pub(crate) data_client: ScsClient<InterceptedService<Channel, HeaderInterceptor>>,
+    pub(crate) data_clients: Vec<ScsClient<InterceptedService<Channel, HeaderInterceptor>>>,
     pub(crate) control_client: ScsControlClient<InterceptedService<Channel, HeaderInterceptor>>,
     pub(crate) configuration: Configuration,
     pub(crate) item_default_ttl: Duration,
 }
 
+static NEXT_DATA_CLIENT_INDEX: AtomicUsize = AtomicUsize::new(0);
```
cprice404, line 86: @kvcache let me know if there's a better way to do this.

sdk/src/cache/cache_client.rs
```diff
@@ -1051,10 +1054,10 @@ impl CacheClient {
     /// * `sorted_set_name` - The name of the sorted set to add an element to.
     /// * `order` - The order to sort the elements by. [SortedSetOrder::Ascending] or [SortedSetOrder::Descending].
     /// * `start_rank` - The rank of the first element to fetch. Defaults to 0. This rank is
-    /// inclusive, i.e. the element at this rank will be fetched.
+    ///   inclusive, i.e. the element at this rank will be fetched.
```
cprice404, line 1057: apologies for these docstring whitespace changes, they were necessary because apparently some clippy rules have been updated since the last time we made a commit to this repo.

sdk/src/errors.rs
```diff
@@ -225,9 +225,9 @@ pub(crate) fn status_to_error(status: tonic::Status) -> MomentoError {
                                 inner_error: Some(status.clone().into()),
                                 details: Some(status.into())
                             },
-                            unknown_err => MomentoError {
-                                message: format!("Unknown error has occurred, unsupported error type for NotFound: {:?}", unknown_err),
-                                error_code: MomentoErrorCode::UnknownError,
+                            _ => MomentoError {
```
cprice404, line 228: This was necessary because the server has changed the err metadata since last time we committed to this SDK. Several tests were failing that were expecting CacheNotFoundError for this case; this fixes them.

sdk/src/cache/cache_client.rs
kvc0: There are other ways but it probably doesn’t matter for what you’re doing here

sdk/src/cache/cache_client.rs
```diff
@@ -76,12 +77,14 @@ use crate::{utils, IntoBytes, MomentoResult};
 /// ```
 #[derive(Clone, Debug)]
 pub struct CacheClient {
-    pub(crate) data_client: ScsClient<InterceptedService<Channel, HeaderInterceptor>>,
+    pub(crate) data_clients: Vec<ScsClient<InterceptedService<Channel, HeaderInterceptor>>>,
```
kvc0, line 80: I realize this is not where pub(crate) started, but it seems more 🤨 with this as a vec property

sdk/src/config/grpc_configuration.rs
```diff
@@ -6,6 +6,8 @@ pub struct GrpcConfiguration {
     /// The duration the client is willing to wait for an RPC to complete before it is terminated
     /// with a DeadlineExceeded error.
     pub(crate) deadline: Duration,
+    /// The number of grpc channels (TCP connections) to create
+    pub(crate) num_channels: u32,
```
kvc0, line 10: Nit: every other configuration in here uses full words - prefer “number.”
Nit: it is a number, it doesn’t need Hungarian notation to help clarify its type. Consider channel_count instead?

sdk/src/cache/cache_client_builder.rs
```diff
@@ -58,17 +62,44 @@ impl CacheClientBuilder<NeedsCredentialProvider> {
 }
 
 impl CacheClientBuilder<ReadyToBuild> {
+    pub fn with_num_connections(self, num_connections: u32) -> CacheClientBuilder<ReadyToBuild> {
```
kvc0, line 65: The usual type for a size is usize.

sdk/src/cache/cache_client_builder.rs
```diff
@@ -58,17 +62,44 @@ impl CacheClientBuilder<NeedsCredentialProvider> {
...
+        let grpc_configuration = self.0.configuration.transport_strategy.grpc_configuration;
+        let transport_strategy = TransportStrategy {
+            grpc_configuration: GrpcConfiguration {
+                num_channels: num_connections,
```
kvc0, line 69: Should this .clamp(1, 32) or something?

kvc0 (commented):
Approach looks fine for what it’s doing. Ideally this would probably be a layer below the client rather than above but I don’t think that would be an api change in the future if it became an interesting optimization

sdk/src/cache/cache_client_builder.rs
cprice404: This is a public-facing API, though... would a user want/expect to know that this is a size? As opposed to just knowing the range of integers that is valid? I don't mind changing it if you think it's the right thing to do but as a rust n00b `u32` seems like a fairly expected/reasonable type to see in this fn signature, while `usize` makes me raise my eyebrows and ask questions about implementation details.

sdk/src/cache/cache_client_builder.rs
cprice404: should it? I can imagine some hard-core scenarios where someone might want like 100 connections?

Also, this kind of relates to the public-facing signature question that I raised above. If we really want to restrict this I would guess that a better UX would be to make that argument a `u8` or something. And if we wanted to restrict it to < 255 then I would expect us to do some validation at construction time and then panic if they pass too high of a value in. clamping seems like a surprising behavior for an end user to experience.

sdk/src/config/grpc_configuration.rs
cprice404: this is an internal variable so if you prefer me to change it i will. in a vacuum I don't necessarily disagree with your naming preference, but in practice, my choice of names here were intended to align with the names of these fields in all of the other existing SDKs that already have support for this, and I probably have a slight pref for consistency with them vs. other naming considerations that only take into account this particular SDK. thoughts?

kvc0: if this is what was done in other sdks then consistency would be the right choice

sdk/src/cache/cache_client_builder.rs
kvc0: usize is commonly used for [container lengths](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.len) in rust, and a connection count sounds like a container limit. u32 is not a bad choice; it supports well and truly more than the range you need, it's just not what I'd expect to see.

cprice404: okay, i can change it to usize if you think that's what a rust dev would expect the API to surface.

sdk/src/cache/cache_client_builder.rs
kvc0: whatever the strategy, it seems worth helping users make a cromulent client rather than allowing clients with 0 channels, resulting in an index out of bound panic later on, or clients with unreasonably high connection counts. Someone can obviously bypass it, but probably the sdk on main should notice if someone tries to make a client with 4294967295 connections.

sdk/src/cache/cache_client_builder.rs
cprice404: done in https://github.com/momentohq/client-sdk-rust/pull/393/commits/9612858e020df7ad4e5c1b6e602b756266efbd59

sdk/src/cache/cache_client_builder.rs
cprice404: fixed in: https://github.com/momentohq/client-sdk-rust/pull/393/commits/c90273feb5bc3b6124e95da54f24d15a44a0bed2

kvc0 (approved):
looks good to me.

anitarua (approved):
lgtm!

sdk/src/cache/cache_client.rs
malandis: For my benefit, you mean that this should be private and if we need an accessor function we should provide one, as opposed to `pub(crate)` which is read/write, correct?

PR #398, opened 2024-11-05
chore: improve resource exhausted message

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1040

When a RESOURCE_EXHAUSTED error is received, we now parse the err metadata in order to report the cause in the message wrapper. If the metadata is not available, we try to use string matching on the error details.

### DIFF
--- a/sdk/src/errors.rs
+++ b/sdk/src/errors.rs
@@ -323,27 +323,38 @@ pub(crate) fn status_to_error(status: tonic::Status) -> MomentoError {
     }
 }
 
-
 enum LimitExceededMessageWrapper {
-    TopicSubscriptionsLimitExceeded,
-    OperationsRateLimitExceeded,
-    ThroughputRateLimitExceeded,
-    RequestSizeLimitExceeded,
-    ItemSizeLimitExceeded,
-    ElementSizeLimitExceeded,
-    UnknownLimitExceeded,
+    TopicSubscriptions,
+    OperationsRate,
+    ThroughputRate,
+    RequestSize,
+    ItemSize,
+    ElementSize,
+    Unknown,
 }
 
 impl LimitExceededMessageWrapper {
     pub fn value(&self) -> String {
         match self {
-            LimitExceededMessageWrapper::TopicSubscriptionsLimitExceeded => "Topic subscriptions limit exceeded for this account".to_string(),
-            LimitExceededMessageWrapper::OperationsRateLimitExceeded => "Request rate limit exceeded for this account".to_string(),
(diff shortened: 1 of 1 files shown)


sdk/src/errors.rs
```diff
@@ -321,3 +322,82 @@ pub(crate) fn status_to_error(status: tonic::Status) -> MomentoError {
...
+    ItemSize,
+    ElementSize,
+    Unknown,
+}
+
+impl LimitExceededMessageWrapper {
+    pub fn value(&self) -> String {
+        match self {
+            LimitExceededMessageWrapper::TopicSubscriptions => {
+                "Topic subscriptions limit exceeded for this account".to_string()
```
malandis, line 340: Could also have this return `&static str` so then no need for `.to_string()` on the variants

sdk/src/errors.rs
```diff
@@ -321,3 +322,82 @@ pub(crate) fn status_to_error(status: tonic::Status) -> MomentoError {
...
+        LimitExceededMessageWrapper::ThroughputRate.value()
+    } else if lower_cased_message.contains("request limit") {
+        LimitExceededMessageWrapper::RequestSize.value()
+    } else if lower_cased_message.contains("item size") {
+        LimitExceededMessageWrapper::ItemSize.value()
+    } else if lower_cased_message.contains("element size") {
+        LimitExceededMessageWrapper::ElementSize.value()
+    } else {
+        LimitExceededMessageWrapper::Unknown.value()
+    }
```
malandis, line 402: Could also simplify this by 
```suggestion
    let wrapper = if lower_cased_message.contains("subscribers") {
        LimitExceededMessageWrapper::TopicSubscriptions
    } else if lower_cased_message.contains("operations") {
        LimitExceededMessageWrapper::OperationsRate
    } else if lower_cased_message.contains("throughput") {
        LimitExceededMessageWrapper::ThroughputRate.value()
    } else if lower_cased_message.contains("request limit") {
        LimitExceededMessageWrapper::RequestSize
    } else if lower_cased_message.contains("item size") {
        LimitExceededMessageWrapper::ItemSize
    } else if lower_cased_message.contains("element size") {
        LimitExceededMessageWrapper::ElementSize
    } else {
        LimitExceededMessageWrapper::Unknown
    }
    wrapper.value()
```

malandis (commented):
No blockers but some style ideas 👍

malandis (dismissed):

malandis (approved):

PR #405, opened 2024-12-09
chore: sdk/Cargo.lock should use correct version of momento packages

malandis:
My local build automatically updates this in the lock file. I suspect
this wasn't committed previously. We will have to monitor after the
next release.


rishtigupta (approved):
:shipit:

PR #406, opened 2024-12-10
feat: add get_batch and set_batch

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/952

### DIFF
--- a/sdk/src/cache/cache_client.rs
+++ b/sdk/src/cache/cache_client.rs
@@ -410,7 +410,7 @@ impl CacheClient {
     /// # cache_client.set(&cache_name, "key1", "value1").await?;
     /// # cache_client.set(&cache_name, "key2", "value2").await?;
     ///
-    /// let results_map: HashMap<String, GetResponse> = cache_client.get_batch(&cache_name, vec!["key1", "key2"]).await?.into();
+    /// let results_map: HashMap<String, GetResponse> = cache_client.get_batch(&cache_name, vec!["key1", "key2"]).await?.try_into().expect("stored string keys");
     /// # assert_eq!(results_map.clone().len(), 2);
     ///
     /// for (key, response) in results_map {
--- a/sdk/src/cache/cache_client.rs
+++ b/sdk/src/cache/cache_client.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::convert::TryInto;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time::Duration;
@@ -26,12 +27,12 @@ use crate::cache::{
     ListPopBackResponse, ListPopFrontRequest, ListPopFrontResponse, ListPushBackRequest,
     ListPushBackResponse, ListPushFrontRequest, ListPushFrontResponse, ListRemoveValueRequest,
     ListRemoveValueResponse, MomentoRequest, SetAddElementsRequest, SetAddElementsResponse,
-    SetFetchRequest, SetFetchResponse, SetIfAbsentOrEqualRequest, SetIfAbsentOrEqualResponse,
-    SetIfAbsentRequest, SetIfAbsentResponse, SetIfEqualRequest, SetIfEqualResponse,
-    SetIfNotEqualRequest, SetIfNotEqualResponse, SetIfPresentAndNotEqualRequest,
-    SetIfPresentAndNotEqualResponse, SetIfPresentRequest, SetIfPresentResponse,
-    SetRemoveElementsRequest, SetRemoveElementsResponse, SetRequest, SetResponse,
-    SortedSetFetchByRankRequest, SortedSetFetchByScoreRequest, SortedSetFetchResponse,
+    SetBatchRequest, SetBatchResponse, SetFetchRequest, SetFetchResponse,
+    SetIfAbsentOrEqualRequest, SetIfAbsentOrEqualResponse, SetIfAbsentRequest, SetIfAbsentResponse,
+    SetIfEqualRequest, SetIfEqualResponse, SetIfNotEqualRequest, SetIfNotEqualResponse,
+    SetIfPresentAndNotEqualRequest, SetIfPresentAndNotEqualResponse, SetIfPresentRequest,
+    SetIfPresentResponse, SetRemoveElementsRequest, SetRemoveElementsResponse, SetRequest,
+    SetResponse, SortedSetFetchByRankRequest, SortedSetFetchByScoreRequest, SortedSetFetchResponse,
     SortedSetGetRankRequest, SortedSetGetRankResponse, SortedSetGetScoreRequest,
     SortedSetGetScoreResponse, SortedSetGetScoresRequest, SortedSetGetScoresResponse,
     SortedSetLengthRequest, SortedSetLengthResponse, SortedSetOrder, SortedSetPutElementRequest,
@@ -305,7 +306,51 @@ impl CacheClient {
         request.send(self).await
     }
 
-    // TODO: set batch
+    /// Sets a batch of items in a Momento Cache
+    ///
+    /// # Arguments
+    ///
+    /// * `cache_name` - name of cache
+    /// * `items` - HashMap of items to set
+    ///
+    /// # Optional Arguments
+    /// If you use [send_request](CacheClient::send_request) to set an item using a
+    /// [SetBatchRequest], you can also provide the following optional arguments:
+    ///
+    /// * `ttl` - The time-to-live for the items. If not provided, the client's default time-to-live is used.
+    ///
+    /// # Examples
+    /// Assumes that a CacheClient named `cache_client` has been created and is available.
+    /// ```
+    /// # fn main() -> anyhow::Result<()> {
+    /// # use momento_test_util::create_doctest_cache_client;
+    /// # tokio_test::block_on(async {
+    /// # let (cache_client, cache_name) = create_doctest_cache_client();
+    /// use std::collections::HashMap;
+    /// use momento::cache::SetResponse;
+    ///
+    /// let items = HashMap::from([("k1", "v1"), ("k2", "v2"), ("k3", "v3")]);
+    /// let results_map: HashMap<String, SetResponse> = cache_client.set_batch(&cache_name, items).await?.into();
+    /// # assert_eq!(results_map.clone().len(), 3);
+    ///
+    /// for (key, response) in results_map {
+    ///     println!("SetResponse for key {}: {:?}", key, response);
+    /// }
+    /// # Ok(())
+    /// # })
+    /// # }
+    /// ```
+    /// You can also use the [send_request](CacheClient::send_request) method to get an item using a [SetBatchRequest].
+    ///
+    /// For more examples of handling the response, see [SetBatchResponse].
+    pub async fn set_batch(
+        &self,
+        cache_name: impl Into<String>,
+        items: HashMap<impl IntoBytes + Copy, impl IntoBytes + Copy>,
+    ) -> MomentoResult<SetBatchResponse> {
+        let request = SetBatchRequest::new(cache_name, items);
+        request.send(self).await
+    }
 
     /// Gets an item from a Momento Cache
     ///
--- a/sdk/src/cache/cache_client.rs
+++ b/sdk/src/cache/cache_client.rs
@@ -16,27 +16,28 @@ use crate::cache::{
     DictionaryLengthResponse, DictionaryRemoveFieldRequest, DictionaryRemoveFieldResponse,
     DictionaryRemoveFieldsRequest, DictionaryRemoveFieldsResponse, DictionarySetFieldRequest,
     DictionarySetFieldResponse, DictionarySetFieldsRequest, DictionarySetFieldsResponse,
-    FlushCacheRequest, FlushCacheResponse, GetRequest, GetResponse, IncreaseTtlRequest,
-    IncreaseTtlResponse, IncrementRequest, IncrementResponse, IntoDictionaryFieldValuePairs,
-    IntoSortedSetElements, ItemGetTtlRequest, ItemGetTtlResponse, ItemGetTypeRequest,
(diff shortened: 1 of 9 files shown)


sdk/src/cache/messages/data/scalar/set_batch.rs
```diff
@@ -0,0 +1,159 @@
...
+        for (key, value) in self.items.iter() {
+            let byte_key = key.into_bytes();
+            let set_request = momento_protos::cache_client::SetRequest {
+                cache_key: byte_key.clone(),
+                cache_body: value.into_bytes(),
+                ttl_milliseconds: cache_client.expand_ttl_ms(self.ttl)?,
+            };
+            set_requests.push(set_request);
+            set_request_keys.push(parse_string(byte_key.clone())?);
+        }
```
anitarua, line 93: fyi: the `<K: IntoBytes + Copy, V: IntoBytes + Copy>` was to address some borrowing issues in this section where the vectors of keys and `SetRequest`s are constructed

sdk/src/cache/messages/data/scalar/set.rs
```diff
@@ -85,11 +87,21 @@ impl<K: IntoBytes, V: IntoBytes> MomentoRequest for SetRequest<K, V> {
...
+            .next_data_client()
+            .set(request)
+            .await?
+            .into_inner();
+        match response.result() {
+            ECacheResult::Ok => Ok(SetResponse {}),
+            _ => Err(MomentoError::unknown_error(
+                "Set",
+                Some(format!("{:#?}", response)),
+            )),
```
anitarua, line 100: previously, the `set` command didn't check for errors like the other sdks do, so this was added to be more consistent

sdk/src/cache/cache_client.rs
```diff
@@ -304,6 +306,52 @@ impl CacheClient {
...
+    /// # })
+    /// # }
+    /// ```
+    /// You can also use the [send_request](CacheClient::send_request) method to get an item using a [SetBatchRequest].
+    ///
+    /// For more examples of handling the response, see [SetBatchResponse].
+    pub async fn set_batch(
+        &self,
+        cache_name: impl Into<String>,
+        items: HashMap<impl IntoBytes + Copy, impl IntoBytes + Copy>,
```
anitarua, line 349: Opted to accept a `HashMap` of (key, value) pairs directly, but could switch to accepting a vector of pairs instead if that seems more user-friendly. 

From what I saw in the other SDKs, set_batch seemed split between a vector of pairs, a dictionary of pairs, or a type union of both.

sdk/src/cache/cache_client.rs
```diff
@@ -343,6 +391,54 @@ impl CacheClient {
...
+    /// # use momento_test_util::create_doctest_cache_client;
+    /// # tokio_test::block_on(async {
+    /// # let (cache_client, cache_name) = create_doctest_cache_client();
+    /// use std::convert::TryInto;
+    /// use std::collections::HashMap;
+    /// use momento::cache::GetResponse;
+    /// # cache_client.set(&cache_name, "key1", "value1").await?;
+    /// # cache_client.set(&cache_name, "key2", "value2").await?;
+    ///
+    /// let results_map: HashMap<String, GetResponse> = cache_client.get_batch(&cache_name, vec!["key1", "key2"]).await?.into();
```
anitarua, line 414: the `get_batch` examples get results as `HashMap<String, GetResponse>` but I can change them to another type if there's another one that may be more commonly accessed

sdk/src/cache/cache_client.rs
kvcache: `+ Copy` is a very aggressive bound. I do not think you want it.
```rust
fn a(aa: impl Sized + Copy) {}

fn b() {
    // works be cause &'static str is Copy
    a("");
    let payload: Vec<u8> = Vec::new();
    // does not compile because Vec is not Copy
    a(payload)
}
```

anitarua: okay, good to know, thanks! I'll look into how to avoid that

kvcache: Similarly, you might find the ergonomics better if you use a looser bound than an explicit HashMap. That does not mean you should accept a vector of pairs either - instead, consider IntoIterator which will let people call you however they want:

```rust
fn c(aa: impl IntoIterator<Item = (&'static str, &'static str)>) {
    let my_proto_map: HashMap<_, _> = HashMap::from_iter(aa);
}

fn d() {
    c([("k1", "v1"), ("k2", "v2")]);

    let mut map = HashMap::new();
    map.insert("k1", "v1");
    map.insert("k2", "v2");
    c(map);

    let mut vec = vec![("k1", "v1"), ("k2", "v2")];
    c(vec);
}
```

sdk/src/cache/messages/data/scalar/get_batch.rs
```diff
@@ -0,0 +1,221 @@
...
+///
+/// let result_map_byte_values: HashMap<String, String> = cache_client.get_batch(&cache_name, keys.clone()).await?.try_into().expect("expecting strings");
+/// # Ok(())
+/// # })
+/// # }
+/// ```
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct GetBatchResponse {
+    results: Vec<GetResponse>,
+    results_dictionary: HashMap<String, GetResponse>,
```
malandis, line 166: My preference would be to store only one of these and build the other on the fly. Ie either only a `Vec` or a `HashMap`.

sdk/src/cache/messages/data/scalar/set.rs
```diff
@@ -1,14 +1,16 @@
...
+use crate::{CacheClient, MomentoError};
 use crate::{IntoBytes, MomentoResult};
 use std::time::Duration;
 
 /// Request to set a value in a cache.
 ///
 /// # Arguments
 ///
-/// * `cache_name` - The name of the cache to create.
+/// * `cache_name` - name of the cache
```
malandis, line 13: Nice find

sdk/src/cache/messages/data/scalar/get_batch.rs
```diff
@@ -0,0 +1,224 @@
...
+        // convert keys to strings for the HashMap keys
+        let string_keys: Vec<String> = byte_keys
+            .iter()
+            .map(|key| parse_string(key.clone()))
+            .collect::<MomentoResult<Vec<String>>>()?;
+
+        let get_requests = byte_keys
+            .iter()
+            .map(|key| momento_protos::cache_client::GetRequest {
+                cache_key: key.clone(),
```
kvcache, line 83: you don't need a memcpy here for these strings
```suggestion
            .into_iter()
            .map(|key| momento_protos::cache_client::GetRequest {
                cache_key: key,
```

sdk/src/cache/messages/data/scalar/get_batch.rs
```diff
@@ -0,0 +1,224 @@
...
+            let key = match string_keys_iter.next() {
+                Some(key) => key,
+                None => {
+                    return Err(MomentoError::unknown_error(
+                        "GetBatch",
+                        Some("Received more responses than expected".to_string()),
+                    ))
+                }
+            };
+            responses.insert(key.to_string(), sdk_get_response);
```
kvcache, line 129: similarly you do not need to memcpy each of the keys again:
```suggestion
        let mut string_keys_iter = string_keys.into_iter();
        while let Some(get_response) = response_stream.message().await? {
            let sdk_get_response = match get_response.result() {
                ECacheResult::Hit => GetResponse::Hit {
                    value: Value {
                        raw_item: get_response.cache_body,
                    },
                },
                ECacheResult::Miss => GetResponse::Miss,
                _ => {
                    return Err(MomentoError::unknown_error(
                        "GetBatch",
                        Some(format!("{:#?}", get_response)),
                    ))
                }
            };
            let key = match string_keys_iter.next() {
                Some(key) => key,
                None => {
                    return Err(MomentoError::unknown_error(
                        "GetBatch",
                        Some("Received more responses than expected".to_string()),
                    ))
                }
            };
            responses.insert(key, sdk_get_response);
```

sdk/src/cache/messages/data/scalar/set_batch.rs
```diff
@@ -0,0 +1,158 @@
...
+
+impl<K: IntoBytes + Copy, V: IntoBytes + Copy> MomentoRequest for SetBatchRequest<K, V> {
+    type Response = SetBatchResponse;
+
+    async fn send(self, cache_client: &CacheClient) -> MomentoResult<SetBatchResponse> {
+        // Turn map of items into a vector of keys and vector of SetRequest objects
+        // so we can map keys to the correct SetResponse objects later
+        let mut set_requests: Vec<momento_protos::cache_client::SetRequest> = vec![];
+        let mut set_request_keys: Vec<String> = vec![];
+        for (key, value) in self.items.iter() {
```
kvcache, line 84: similarly, this should be using into_iter() so that the into_bytes implementation can use the owned implementation if there is one available, potentially skipping a whole layer of memcpy for the batch

sdk/src/cache/messages/data/scalar/set_batch.rs
```diff
@@ -0,0 +1,158 @@
...
+            let byte_key = key.into_bytes();
+            let set_request = momento_protos::cache_client::SetRequest {
+                cache_key: byte_key.clone(),
+                cache_body: value.into_bytes(),
+                ttl_milliseconds: cache_client.expand_ttl_ms(self.ttl)?,
+            };
+            set_requests.push(set_request);
+            set_request_keys.push(parse_string(byte_key.clone())?);
```
kvcache, line 92: unnecessary clone - clippy should catch this :-(
```suggestion
            set_request_keys.push(parse_string(byte_key)?);
```

sdk/src/cache/messages/data/scalar/set_batch.rs
```diff
@@ -0,0 +1,158 @@
...
+
+        let mut response_stream = cache_client
+            .next_data_client()
+            .set_batch(set_batch_request)
+            .await?
+            .into_inner();
+
+        // receive stream of get responses
+        let mut responses: HashMap<String, SetResponse> = HashMap::new();
+        let mut set_request_keys_iter = set_request_keys.iter();
```
kvcache, line 111: just like in the get batch side, this should be into_iter, and you should not to_string() on line 131 - to avoid extra O(n) memcpys

kvcache (commented):
a few little internal notes I noticed. Might help you with the `Copy` thing, not sure

sdk/src/cache/messages/data/scalar/get_batch.rs
anitarua: I opted to remove the `Vec<GetResponse>` and `Vec<SetResponse>` options instead. `HashMap.iter()` will traverse the items in an arbitrary order, not the order in which the batch of items was received. I think having a direct mapping from key <-> response is clearer.

sdk/src/cache/messages/data/scalar/set_batch.rs
anitarua: the `iter()` -> `into_iter()` changes helped a lot with removing the need for `+ Copy`. Thanks @kvcache !

kvcache (approved):
Code looks good to me

sdk/src/cache/messages/data/scalar/get_batch.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    async fn send(self, cache_client: &CacheClient) -> MomentoResult<GetBatchResponse> {
+        // consume self.keys once to convert all keys to bytes
+        let byte_keys: Vec<Vec<u8>> = self.keys.into_bytes();
+
+        // convert keys to strings for the HashMap keys
+        let string_keys: Vec<String> = byte_keys
+            .iter()
+            .map(|key| parse_string(key.clone()))
+            .collect::<MomentoResult<Vec<String>>>()?;
```
malandis, line 78: Isn't this a problem if the keys are actually bytes that can't be represented as utf-8? The caller can pass in `IntoBytes` to represent the keys, which can be `Vec<u8>` or various string types. In those cases this will fail.

To guard against that we should store the keys as bytes and convert in the `try_from`s below.

sdk/src/cache/messages/data/scalar/set_batch.rs
```diff
@@ -0,0 +1,152 @@
...
+        let mut set_request_keys: Vec<String> = vec![];
+        for (key, value) in self.items.into_iter() {
+            let byte_key = key.into_bytes();
+            let set_request = momento_protos::cache_client::SetRequest {
+                cache_key: byte_key.clone(),
+                cache_body: value.into_bytes(),
+                ttl_milliseconds: cache_client.expand_ttl_ms(self.ttl)?,
+            };
+            set_requests.push(set_request);
+            set_request_keys.push(parse_string(byte_key)?);
```
malandis, line 92: I have the same concern here re: interpreting the keys as strings. This will fail if the keys are binary that can't be converted to utf-8.

sdk/src/cache/messages/data/scalar/get_batch.rs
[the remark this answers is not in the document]
malandis: JavaScript allows byte arrays as the keys, see: [[1]](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/cache-data-client.ts#L1860) and [[2]](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/batchutils/batch-props.ts#L16-L20). We should store this as bytes and convert as needed.

malandis (commented):
Looks good. Last couple recommendations

- Let's store keys as bytes (`Vec<u8>`) and converting in the responses. Since we allow the caller to pass in binary keys, they may pass in data that is not representable as utf-8. Because of that we should only make that conversion upon request. We do this in the JS SDK.
- Let's add some tests to cover passing in bytes as keys/values.

malandis (approved):

PR #418, opened 2024-12-18
chore: showcase publisher_id in topics example

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/951

Now that generate_disposable_token is implemented, I updated this example to include that as well.

### DIFF
--- a/example/rust/src/bin/topics.rs
+++ b/example/rust/src/bin/topics.rs
@@ -16,9 +16,15 @@ async fn main() -> MomentoResult<()> {
     // call `abort()` on the task handle after messages are published.
     let mut subscription1 = topic_client.subscribe("cache", "my-topic").await?;
     let subscriber_handle1 = tokio::spawn(async move {
-        println!("Subscriber should keep receiving until task is aborted");
+        println!("\n Subscriber [1] should keep receiving until task is aborted");
         while let Some(message) = subscription1.next().await {
-            println!("[1] Received message: {:?}", message);
+            println!(
+                "[1] Received message: \n\tKind: {:?} \n\tSequence number: {:?} \n\tSequence page: {:?} \n\tPublisher ID: {:?}", 
+                message.kind,
+                message.topic_sequence_number,
+                message.topic_sequence_page,
+                message.publisher_id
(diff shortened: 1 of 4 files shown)


example/rust/src/bin/topics.rs
```diff
@@ -38,10 +49,24 @@ async fn main() -> MomentoResult<()> {
     // let the task end after receiving 10 messages.
     let mut subscription2 = topic_client.subscribe("cache", "my-topic").await?;
     tokio::spawn(async move {
-        println!("Subscriber should receive 10 messages then exit");
+        println!("\nSubscriber [2] should receive 10 messages then exit");
         for _ in 0..10 {
             let message = subscription2.next().await;
-            println!("[2] Received message: {:?}", message);
+            // println!("[2] Received message: {:?}", message);
```
malandis, line 55: Can delete this line.

example/rust/src/bin/topics.rs
```diff
@@ -38,10 +49,24 @@ async fn main() -> MomentoResult<()> {
     // let the task end after receiving 10 messages.
     let mut subscription2 = topic_client.subscribe("cache", "my-topic").await?;
     tokio::spawn(async move {
-        println!("Subscriber should receive 10 messages then exit");
+        println!("\nSubscriber [2] should receive 10 messages then exit");
```
malandis, line 52: For example 2 it would be clearer to use synchronization to detect 10 messages are received. Currently the code relies on delays and the counter. Something like tokio's oneshot channel could work here [[1]](https://tokio.rs/tokio/tutorial/channels) [[2]](https://docs.rs/tokio/1.42.0/tokio/sync/oneshot/index.html)

malandis (commented):
Example 2 could use some rework but looks good otherwise.

example/rust/src/bin/topics.rs
anitarua: good suggestion! I ended up using `mpsc` to send multiple messages instead of `oneshot`

pgautier404 (approved):

Issue #419, opened 2024-12-18
Support generating disposable tokens


danielamiao:
One suggestion is to have a builder with the required input values:

```
ExpiresIn utils.ExpiresIn
Scope     DisposableTokenScope
Props     DisposableTokenProps
```

anitarua:
- [ ] create auth client and permissions structure too since that doesn't exist yet in Rust sdk
- [ ] add generateDisposableToken API

PR #420, opened 2024-12-19
feat: add auth client and generate_disposable_token api

anitarua:
Work towards https://github.com/momentohq/client-sdk-rust/issues/419

- Implemented AuthClient and builder. No config object since there's no logging object to configure in this SDK like in the JS or Go ones.
- Implemented permissions structure (`PermissionScope` and `DisposableTokenScope` and related objects)
- Implemented `generate_disposable_token` method
- Added docs and doctests for public-facing objects
- Added integration tests to verify correctness of `generate_disposable_token` implementation
- Updated test-util and on-pull-request workflow to accommodate new tests

After cutting a release, I can fast-follow with updating the docs snippets example file and unit tests for the permissions conversions.

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -9,6 +9,7 @@ env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   TEST_CACHE_NAME: rust-integration-test-ci-cache-${{ github.sha }}
   TEST_STORE_NAME: rust-integration-test-ci-store-${{ github.sha }}
+  TEST_AUTH_CACHE_NAME: rust-integration-test-ci-cache-auth-${{ github.sha }}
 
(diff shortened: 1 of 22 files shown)


benbpyle:
Thanks for implementing this feature!

honeyAcorn (changes requested):

honeyAcorn:
I've verified the capability by using momento console to generate superuser api key.
Then use the Superuser API key to generate disposable token using rust sdk.
Those disposable token can only be used toward certain cache for only read or write or both.
Everything looks good except for some missing test case.
1. prefix-write
2. Can we add a verify for the expire time? Add one case to expire in a few seconds and test if it expires in integration test.

honeyAcorn (approved):
LGTM!

pgautier404 (approved):

PR #427, opened 2025-01-04
fix: add permission scope unit tests and From conversions for convenience

anitarua:
Follow up for https://github.com/momentohq/client-sdk-rust/issues/419

Adds more unit tests for the permissions structure for the auth client to match the test coverage we have in the JS sdk.

In the process of writing these tests, I also:

- Wrote more `From` implementations to make conversion between relevant objects easier and refactored the disposable tokens integration tests to make use of these.
- Added `#[derive(Debug, Display, Clone, PartialEq)]` statements to permissions objects to make more `assert`s possible.
- Made CacheItemKey and CacheItemKeyPrefix store `Vec<u8>` directly instead of `<impl IntoBytes>` so that we can avoid the propagation of the generic type everywhere

### DIFF
--- a/src/auth/messages/permissions_conversions.rs
+++ b/src/auth/messages/permissions_conversions.rs
@@ -24,7 +24,7 @@ use crate::{
 
 // Create protobuf Permissions from DisposableTokenScope
 pub(crate) fn permissions_from_disposable_token_scope(
-    scope: DisposableTokenScope<impl IntoBytes>,
+    scope: DisposableTokenScope,
 ) -> permission_messages::Permissions {
     match scope {
         DisposableTokenScope::Permissions(permissions) => permission_messages::Permissions {
@@ -132,9 +132,7 @@ fn topic_permission_to_grpc_permission(permission: TopicPermission) -> Permissio
     }
 }
 
-fn assign_cache_item_selector(
-    item_selector: CacheItemSelector<impl IntoBytes>,
-) -> cache_permissions::CacheItem {
+fn assign_cache_item_selector(item_selector: CacheItemSelector) -> cache_permissions::CacheItem {
     match item_selector {
         CacheItemSelector::AllCacheItems => cache_permissions::CacheItem::AllItems(All {}),
         CacheItemSelector::CacheItemKey(CacheItemKey { key }) => {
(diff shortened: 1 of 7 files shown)


pgautier404 (approved):
LGTM

src/auth/permissions/permission_scope.rs
```diff
@@ -38,6 +40,7 @@ impl From<&str> for CacheSelector {
 }
 
 /// Defines access permissions for a cache.
+#[derive(Debug, Clone, PartialEq)]
```
malandis, line 43: You'll also want `Display` on these (here and elsewhere).

src/auth/messages/permissions_conversions.rs
```diff
@@ -164,3 +164,291 @@ fn disposable_token_permission_to_grpc_permission(
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::auth::{DisposableTokenCachePermissions, DisposableTokenScope, Permissions};
+
+    #[test]
+    fn creates_expected_grpc_permissions_from_all_data_read_write() {
+        let sdk_permissions =
+            DisposableTokenScope::Permissions::<String>(Permissions::all_data_read_write());
```
malandis, line 176: Why do we need `<String>` here?

anitarua: this resulted from implementing the [CacheItemKey and CacheItemKeyPrefix structs](https://github.com/momentohq/client-sdk-rust/blob/main/src/auth/permissions/disposable_token_scope.rs#L6-L29) with the ability to accept keys as `impl IntoBytes`

if it looks too weird, we could change it so that keys/prefixes are accepted only as bytes

malandis: OK. I recommend tagging someone on the service team for a Rust ergonomics review. 👍

anitarua: Took Vir's advice and made the key and prefix structs store `Vec<u8>` directly, which allows us to avoid propagating the generic type all the way.

malandis (approved):

malandis:
@virratanasangpunth may want to chime in here as well given he suggested the change.

virratanasangpunth (approved):
LGTM

Issue #429, opened 2025-01-06
PRs from contributors do not run approvals


kvcache:
The approval workflows in this repo do not work with forks. Contributors who want to improve documentation, add features, or fix bugs can't self-service a PR to the point of being reviewable & mergeable.

The issue is with secrets that are accessible to the upstream repository, but not to forks. The PR approvals should not have a dependency on access tokens, or there should at least be some approvals sufficient for mergeability that are runnable without having access to private secrets.

See: https://github.com/momentohq/client-sdk-rust/pull/422

anitarua:
Recently solved this issue in JS sdk, can implement here as well: https://github.com/momentohq/client-sdk-javascript/pull/1485

PR #422, opened 2024-12-28
Example Rust project for subscribing to a Topic

benbpyle:
* Added Main
* Added Model to be serialized
* Added README file
* updated the gitignore

PR #432, opened 2025-01-09
ci: allow maintainers to approve test workflow on PRs from forks

anitarua:
Closes https://github.com/momentohq/client-sdk-rust/issues/429

Currently, PRs from forks cannot successfully run integration tests because they lack access to our api key secret.
This blocks open-source contributors from being able to easily work on our repo (see #422).

This PR implements the same github actions workflow we had implemented in the JS repo to allow maintainers to approve/rerun workflows on PRs from forks and allow them access to the api key secret (see https://github.com/momentohq/client-sdk-javascript/pull/1485).

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -1,7 +1,7 @@
 name: on-pull-request
 
 on:
-  pull_request:
+  pull_request_target:
     branches: [main]
 
 env:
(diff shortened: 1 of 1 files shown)


malandis (commented):
Do we not need to explicitly mention the `ref` anymore in the checkout step?

anitarua:
> Do we not need to explicitly mention the ref anymore in the checkout step?

No, we don't. As far as I understand, we want the default ref so that it checks out the simulated merge commit. 
The ref that was in the article would have checked out the code as it was on the PR I think.

malandis (approved):

PR #437, opened 2025-01-13
ci: checkout default ref on PRs if merge commit is unavailable

anitarua:
It appears that sometimes the `merge_commit_sha` we want to check out in combination with using `on: pull_request_target` does not always exist.

This PR ensures we checkout `merge_commit_sha || pull_request.head.sha` so that it has a fallback ref that will always exist. (The head.sha is the last commit on the PR, the merge_commit_sha is the test merge with main).

Note: The updated workflow will not run on this PR because `pull_request_target` runs in the context of the base branch (main), not the PR branch.

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -35,7 +35,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


rishtigupta (approved):
:shipit:

malandis (approved):

PR #438, opened 2025-01-25
feat: momento leaderboard client

brayniac:
Adds client functionality for the Momento Leaderboard API.


src/leaderboard/leaderboard_resource.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub async fn remove_elements<T: IntoIds>(
+        &self,
+        ids: T,
+    ) -> MomentoResult<RemoveElementsResponse> {
+        let request = RemoveElementsRequest::new(ids);
+        request.send(self).await
+    }
+
+    /// Upsert (update/insert) elements into a leaderboard.
+    pub async fn upsert<E: IntoElements>(&self, elements: E) -> MomentoResult<UpsertResponse> {
```
kvcache, line 82: Similar comments as above.

For the trait though, did you consider not creating IntoElements and instead using the usual Rust `Into` on a single value? You would delete IntoElements, write `impl From<(u32, f64)> for Element` and then use simple functions:
`pub async fn upsert(&self, element: impl Into<Element>)`
`pub async fn upsert_batch(&self, elements: impl IntoIterator<Item = impl Into<Element>>)`

This would also enable your users to more simply implement translators from their own business objects into unary and batch compatible representations.

src/leaderboard/leaderboard_resource.rs
```diff
@@ -0,0 +1,128 @@
...
+        let next_index =
+            NEXT_DATA_CLIENT_INDEX.fetch_add(1, Ordering::Relaxed) % self.data_clients.len();
+        self.data_clients[next_index].clone()
+    }
+
+    pub(crate) fn deadline(&self) -> Duration {
+        self.deadline
+    }
+
+    pub(crate) fn cache_name(&self) -> &String {
```
kvcache, line 114: `&String` is not normal. Unless there's a specific reason you need a String reference you should return `&str` instead (here and elsewhere). Using &String makes it misleading.

src/leaderboard/leaderboard_resource.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    pub(crate) fn leaderboard_name(&self) -> &String {
+        &self.leaderboard_name
+    }
+
+    /// Lower-level API to send any type of MomentoRequest to the server. This is used for cases when
+    /// you want to set optional fields on a request that are not supported by the short-hand API for
+    /// that request type.
+    pub async fn send_request<R: MomentoRequest>(&self, request: R) -> MomentoResult<R::Response> {
```
kvcache, line 125: This trait name is rough, and overloaded with other MomentoRequest. It's going to be hard to read in user code. Consider using something non-overlapping and clear like `LeaderboardRequest` instead.

src/leaderboard/messages/momento_request.rs
```diff
@@ -0,0 +1,17 @@
+use crate::{Leaderboard, MomentoResult};
+
+/// A trait that allows Momento request types to define their interaction with the gRPC client.
+pub trait MomentoRequest {
+    #[allow(missing_docs)]
+    type Response;
```
kvcache, line 6: Why should this allow missing docs? Is there nothing to say about this, but a reason to not hide it from docs?

src/leaderboard/messages/momento_request.rs
```diff
@@ -0,0 +1,17 @@
...
+
+    /// An internal fn that allows Momento request types to define their interaction with
+    /// the gRPC client. You can impl this fn for your own types if you'd like to hand them
+    /// to the Momento client directly, but that is not an explicitly supported scenario and
+    /// this signature may change a little over time. If that's okay with you, impl away!
```
kvcache, line 11: This guidance is untrue, for reasonable interpretations. `next_data_client` is crate-private, and `send` is expected to implement actually sending.

Additionally, this is not how momento should support users plugging in to (currently called) MomentoRequest. Users should `impl From<MyOwnStruct> for LeaderboardFetchRequest` and use `leaderboard.send(my_own.into())` or something to that effect.

src/leaderboard/messages/data/delete.rs
```diff
@@ -0,0 +1,45 @@
...
+        Self::new()
+    }
+}
+
+impl MomentoRequest for DeleteRequest {
+    type Response = DeleteResponse;
+
+    async fn send(self, leaderboard: &Leaderboard) -> MomentoResult<Self::Response> {
+        let cache_name = leaderboard.cache_name();
+        let request = prep_leaderboard_request_with_timeout(
```
kvcache, line 26: yeah, "timeout" word is correct here, but needs fixed 2 lines down

src/leaderboard/messages/data/fetch.rs
```diff
@@ -0,0 +1,20 @@
...
+impl FetchResponse {
+    /// Constructs a new `FetchResponse`.
+    pub fn new(elements: Vec<RankedElement>) -> Self {
+        Self { elements }
+    }
+
+    /// Returns the ranked elements in the response.
+    pub fn elements(&self) -> &[RankedElement] {
+        &self.elements
+    }
```
kvcache, line 19: there should be an `into_elements(self)` as well here to avoid forcing users to clone if they want owned elements.

src/leaderboard/messages/data/fetch_by_rank.rs
```diff
@@ -0,0 +1,88 @@
...
+            rank_range: rank_range.into(),
+            order: Order::Ascending,
+        }
+    }
+
+    /// Sets the order of the elements to be fetched.
+    pub fn order(mut self, order: Order) -> Self {
+        self.order = order;
+        self
+    }
```
kvcache, line 51: somewhere, or in both places, this should specify what the ordering default is

src/leaderboard/messages/data/fetch_by_rank.rs
```diff
@@ -0,0 +1,88 @@
...
+            .next_data_client()
+            .get_by_rank(request)
+            .await?
+            .into_inner();
+
+        Ok(Self::Response::new(
+            response
+                .elements
+                .iter()
+                .map(|v| RankedElement {
```
kvcache, line 80: getrank, fetchbyscore, fetchbyrank - is v the same type in all 3? Just `impl From<Whatever V Is> for RankedElement` and replace this line and the other 2 with `.map(|v| v.into())` instead of repeating the same marshaling stanza

src/leaderboard/messages/data/fetch_by_score.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// Constructs a new `ScoreRange` with no bounds.
+    pub fn unbounded() -> Self {
+        Self {
+            min: None,
+            max: None,
+        }
+    }
+
+    /// Validates the score range.
+    pub fn validate(&self) -> MomentoResult<()> {
```
kvcache, line 32: why is this public? Why is it possible to create an invalid ScoreRange? Shouldn't `new()` return a `MomentoResult<Self>`?

Should `new` just `.clamp` min and max to `f64::MIN` and `f64::MAX` instead of making it an error to write `1.0..`?

src/leaderboard/messages/data/fetch_by_score.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+        Ok(())
+    }
+}
+
+impl From<Option<ScoreRange>> for ScoreRange {
+    fn from(val: Option<ScoreRange>) -> Self {
+        val.unwrap_or_else(ScoreRange::unbounded)
+    }
+}
```
kvcache, line 61: This appears to be unused, and it is unclear to me that unbounded means the same thing as None for a whole score range. If you want to keep it, what's the utility here, and how should this be communicated so it doesn't surprise people that `None` means `Everything`?

kvcache (commented):
I ran out of time today about halfway through. I'll give this feedback before continuing to review tomorrow.

src/leaderboard/config/configuration.rs
```diff
@@ -0,0 +1,88 @@
...
+///     .transport_strategy(
+///         TransportStrategy::builder()
+///             .grpc_configuration(
+///                 GrpcConfiguration::builder()
+///                     .deadline(Duration::from_millis(1000))
+///             )
+///     );
+
+#[derive(Clone, Debug)]
+pub struct Configuration {
```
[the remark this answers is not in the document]
malandis: Do you mean why not just re-use the cache configuration object here?

src/leaderboard/leaderboard_client.rs
```diff
@@ -0,0 +1,48 @@
...
+use crate::leaderboard::Leaderboard;
+
+use momento_protos::leaderboard::leaderboard_client::LeaderboardClient as SLbClient;
+use tonic::codegen::InterceptedService;
+use tonic::transport::Channel;
+
+/// Client to work with Momento Leaderboards.
+#[derive(Clone, Debug)]
+pub struct LeaderboardClient {
+    data_clients: Vec<SLbClient<InterceptedService<Channel, HeaderInterceptor>>>,
```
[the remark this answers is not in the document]
malandis: I've changed this to qualify from the leaderboard proto mod instead as per your suggestion in [5803c70](https://github.com/momentohq/client-sdk-rust/pull/438/commits/5803c70d60f5d9a8414de3e446482cd0528a3321)

src/leaderboard/leaderboard_resource.rs
```diff
@@ -0,0 +1,128 @@
...
+use std::time::Duration;
+
+static NEXT_DATA_CLIENT_INDEX: AtomicUsize = AtomicUsize::new(0);
+
+use super::messages::data::IntoIds;
+
+/// Represents a remote leaderboard resource.
+pub struct Leaderboard {
+    data_clients: Vec<SLbClient<InterceptedService<Channel, HeaderInterceptor>>>,
+    deadline: Duration,
```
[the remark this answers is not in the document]
malandis: Yeah that's right. I was following the naming convention elsewhere in this repo (see [here](https://github.com/momentohq/client-sdk-rust/blob/main/src/config/grpc_configuration.rs#L7-L10)). I'll have to loop back and address that in the cache client and topic clients.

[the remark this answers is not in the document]
malandis: I've renamed this in the leaderboard mod to `client_timeout` which aligns with other SDKs in [ada9c97](https://github.com/momentohq/client-sdk-rust/pull/438/commits/ada9c97b091ddef67c4c48ef951f90e8a4774221).

I've made a note for the backlog to address this on the cache and topics side.

src/leaderboard/leaderboard_resource.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub async fn fetch_by_score(
+        &self,
+        score_range: impl Into<ScoreRange>,
+    ) -> MomentoResult<FetchResponse> {
+        let request = FetchByScoreRequest::new(score_range);
+        request.send(self).await
+    }
+
+    /// Get the length of a leaderboard.
+    pub async fn length(&self) -> MomentoResult<LengthResponse> {
```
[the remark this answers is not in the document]
malandis: I've applied your suggestion in [4f32b66](https://github.com/momentohq/client-sdk-rust/pull/438/commits/4f32b66037adfe7eec90dd9e212ba871ab035a72)

src/leaderboard/messages/data/delete.rs
malandis: Addressed in https://github.com/momentohq/client-sdk-rust/pull/438#discussion_r1943405912

src/leaderboard/leaderboard_resource.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl Leaderboard {
+    /// Delete a leaderboard.
+    pub async fn delete(&self) -> MomentoResult<DeleteResponse> {
+        let request = DeleteRequest::new();
+        request.send(self).await
+    }
+
+    /// Fetch elements from a leaderboard by rank.
+    pub async fn fetch_by_rank(
```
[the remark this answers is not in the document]
malandis: Clarified in [c5800e6](https://github.com/momentohq/client-sdk-rust/pull/438/commits/c5800e6c8a0e9be406c6a1b4abc416c1f6fc1ad1)

src/leaderboard/messages/data/fetch_by_rank.rs
malandis: Clarified in [c5800e6](https://github.com/momentohq/client-sdk-rust/pull/438/commits/c5800e6c8a0e9be406c6a1b4abc416c1f6fc1ad1)

src/leaderboard/leaderboard_resource.rs
malandis: Fixed in [f31495c](https://github.com/momentohq/client-sdk-rust/pull/438/commits/f31495c5d101f527feba5c41137db0c9c0b1e959)

src/leaderboard/leaderboard_resource.rs
```diff
@@ -0,0 +1,128 @@
...
+        &self,
+        rank_range: impl Into<RankRange>,
+    ) -> MomentoResult<FetchResponse> {
+        let request = FetchByRankRequest::new(rank_range);
+        request.send(self).await
+    }
+
+    /// Get elements from a leaderboard by score.
+    pub async fn fetch_by_score(
```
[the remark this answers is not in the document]
malandis: Clarified in [c5800e6](https://github.com/momentohq/client-sdk-rust/pull/438/commits/c5800e6c8a0e9be406c6a1b4abc416c1f6fc1ad1)

src/leaderboard/config/configuration.rs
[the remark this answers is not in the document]
kvcache: I guess there are unrelated concepts in cache, but more configurators seems more complex to reason about. I dunno, it's probably fine this way.

src/leaderboard/leaderboard_resource.rs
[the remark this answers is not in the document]
kvcache: Ah, yeah thank you for that!

src/leaderboard/leaderboard_resource.rs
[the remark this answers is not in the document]
kvcache: perfect

src/leaderboard/messages/momento_request.rs
malandis: I've clarified in [2e155bd](https://github.com/momentohq/client-sdk-rust/pull/438/commits/2e155bd21c33ff9a887f04865841f73664ce968b). I copied this comment from the cache client. Looking at that, I'm not sure why we would to make it pluggable. You're right that unless we make `next_data_client` public, then the end user has no way of sending the request in their implementation.

src/leaderboard/messages/momento_request.rs
malandis: Copied over from the cache `MomentoRequest`. I've clarified in [2e155bd](https://github.com/momentohq/client-sdk-rust/pull/438/commits/2e155bd21c33ff9a887f04865841f73664ce968b).

src/leaderboard/config/configuration.rs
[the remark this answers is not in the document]
malandis: Ack. As time goes on we've been bitten by sharing (at first blush) same configurations/configuration parts across different clients (cache vs topics vs storage vs leaderboard). Once we need to specialize a specific client's configuration, we don't want to affect the other clients. So for forwards compatibility it's safer to separate them.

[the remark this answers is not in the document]
kvcache: Roger that, thank you for the explanation!

src/leaderboard/leaderboard_resource.rs
malandis: Good idea. I updated this in [87dc81a](https://github.com/momentohq/client-sdk-rust/pull/438/commits/87dc81a20d361e0df05e9314f12c55bacb8ac939). We still have a type def for `MomentoResult<T> = Result<T, MomentoError>` which is shared across the clients.

src/leaderboard/messages/data/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn into_ids(self) -> Vec<u32>;
+}
+
+#[cfg(not(doctest))]
+pub(crate) fn map_and_collect_elements<'a, I>(iter: I) -> Vec<u32>
+where
+    I: Iterator<Item = &'a u32>,
+{
+    iter.copied().collect()
+}
```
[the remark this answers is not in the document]
malandis: Removed in [b2a03a9](https://github.com/momentohq/client-sdk-rust/pull/438/commits/b2a03a9eb6bc8344eec290c581502c779b05aac8)

src/leaderboard/messages/data/fetch_by_score.rs
malandis: Good catch -- this was a leftover while I was iterating on the UI. I've removed this in [e930128](https://github.com/momentohq/client-sdk-rust/pull/438/commits/e93012847c0f00f949768003d48bb0eb4815a07c)

src/leaderboard/messages/data/fetch_by_rank.rs
malandis: Nice. Fixed in [305d247](https://github.com/momentohq/client-sdk-rust/pull/438/commits/305d24749db808f5c753ba869fbb3079af0f5614).

src/leaderboard/messages/data/fetch_by_rank.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use std::ops::Range;
+
+/// Represents a range of ranks used to request elements by rank.
+pub struct RankRange {
+    start_inclusive: u32,
+    end_exclusive: u32,
+}
+
+impl From<Range<u32>> for RankRange {
```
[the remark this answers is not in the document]
malandis: Done in [d749551](https://github.com/momentohq/client-sdk-rust/pull/438/commits/d7495519abed88464f3a7edb4b8eb30f0ded3345). I'm unclear how to convey the unlikely but troubling edge case if a user provides `u32::MAX` as the end index inclusive and we convert to exclusive. Any suggestions?

src/leaderboard/messages/data/fetch_by_score.rs
malandis: `1.0..` not working was an oversight my on part. I've added `From<RangeFrom<f64>>` and `From<RangeTo<f64>>` to handle those cases in [0576f7a](https://github.com/momentohq/client-sdk-rust/pull/438/commits/0576f7a470c9655480c712f02611a9d738dbd4f3)

malandis: The error cases are when the user passes in `min=fp64::POS_INFINITY` or `max=fp64::NEG_INFINITY`. We translate those to the proto `Unbounded` message. That's why the check `val.is_finite()` isn't enough before marshalling to `Unbounded`: if `min` or `max` is the wrong infinity, then it's an error.

malandis: Well maybe not. The top level methods take as input `impl Into<ScoreRange>`. If we let `new` return `MomentoResult<ScoreRange>` then the callers will have to invoke the function with something like `client.fetch_by_score((1.0..5.0).into()?)`, right? Because the `From<Range<f64>>` implementation is for `ScoreRange`, not `MomentoResult<ScoreRange>`. See j[1c7c02d](https://github.com/momentohq/client-sdk-rust/pull/438/commits/1c7c02d82472e56e412b304d838ec85404c75b42)

src/leaderboard/leaderboard_resource.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /// Get the length of a leaderboard.
+    pub async fn length(&self) -> MomentoResult<LengthResponse> {
+        let request = LengthRequest::new();
+        request.send(self).await
+    }
+
+    /// Get rank of elements from a leaderboard using their element ids.
+    pub async fn get_rank<T: IntoIds>(&self, ids: T) -> MomentoResult<GetRankResponse> {
```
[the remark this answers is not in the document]
malandis: > What is the order this function counts rank from? If id 1234 is rank 42, do they have the 42nd lowest or 42nd highest score? Is it the 43rd?

I've documented the order now

> In the above functions you used impl Trait syntax, but here and below you use bounded T syntax. Why not use the lighter impl IntoIds here?

Good point. I inherited the code and don't have a good intuition yet as for the right amount of baggage. Will opt for the simpler case here.

> Did you consider not defining IntoIds? You could do this more simply by having pub async fn get_rank(&self, id: u32) and pub async fn get_ranks(&self, ids: impl IntoIterator<u32>) which would allow people to use a wide ecosystem of existing ways to express an iterable of numbers.

I like this better. In the other SDKs apparently there is only one function defined for each. I prefer a little more work on our side to make a lot of dev's life slightly easier. 👍

src/leaderboard/leaderboard_resource.rs
malandis: I've made the change to use `impl IntoIterator` in [981b5b7](https://github.com/momentohq/client-sdk-rust/pull/438/commits/981b5b7a02f5a5bb5acc27bca5ec941914e409d1). I decided to leave this as a single `upsert` function for now to align with the leaderboard pattern in other languages.

src/leaderboard/leaderboard_resource.rs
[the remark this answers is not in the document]
malandis: I broadened the use of `IntoIds` as per your suggestion @ [67da8e0](https://github.com/momentohq/client-sdk-rust/pull/438/commits/67da8e0f75db7a5474c22dfd99007373406cdf8f)

src/leaderboard/config/configuration.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// Low-level options for network interactions with Momento.
+    pub(crate) transport_strategy: TransportStrategy,
+}
+
+impl Configuration {
+    /// First level of constructing a CacheClient configuration. Must provide a [TransportStrategy] to continue.
```
anitarua, line 40:
```suggestion
    /// First level of constructing a LeaderboardClient configuration. Must provide a [TransportStrategy] to continue.
```

src/leaderboard/leaderboard_resource.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub async fn len(&self) -> MomentoResult<LengthResponse> {
+        let request = LengthRequest::new();
+        request.send(self).await
+    }
+
+    /// Get rank of elements from a leaderboard using their element ids.
+    ///
+    /// Defaults to ascending order rank, meaning rank 0 is the element with
+    /// the lowest score.
+    pub async fn get_rank(
```
anitarua, line 80: @nand4011 I think we ran into this question before, but doesn't get_rank (or one of the fetch methods) return elements in order of ID instead of rank or score?

src/leaderboard/leaderboard_resource.rs
```diff
@@ -0,0 +1,152 @@
...
+        &self.cache_name
+    }
+
+    pub(crate) fn leaderboard_name(&self) -> &str {
+        &self.leaderboard_name
+    }
+
+    /// Lower-level API to send any type of MomentoRequest to the server. This is used for cases when
+    /// you want to set optional fields on a request that are not supported by the short-hand API for
+    /// that request type.
```
anitarua, line 145: might be good to link to one of the message files that include an example of using send_request, like we do for cache

src/leaderboard/messages/data/length.rs
```diff
@@ -0,0 +1,58 @@
...
+            .await?
+            .into_inner();
+
+        Ok(Self::Response {
+            length: response.count,
+        })
+    }
+}
+
+/// The response type for a successful `LengthRequest`
```
anitarua, line 47: can be a follow up, but I think we should include examples of using the method and response in the docs comments, like we do for cache

src/leaderboard/config/configuration.rs
```diff
@@ -0,0 +1,88 @@
+use std::time::Duration;
+
+use crate::config::transport_strategy::TransportStrategy;
```
nand4011, line 3: Is it worth using a leaderboard-specific transport strategy and grpc configuration?
We'll need to figure out a general strategy if we need to separate the streaming and unary calls in the rust topic client anyway.

tests/leaderboard/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+            "Expected two elements in the leaderboard"
+        );
+
+        leaderboard.delete().await?;
+        let length_response2 = leaderboard.len().await?;
+
+        assert_eq!(
+            0,
+            length_response2.length(),
+            "Expected two elements in the leaderboard"
```
anitarua, line 82:
```suggestion
            "Expected zero elements in the leaderboard"
```

anitarua (commented):
looking good! just had a couple questions/nits

nand4011 (commented):
I think we need docs examples for making leaderboard calls using send_request, but we can do that in a follow up pr.

Just a config question, but looks good otherwise.

src/leaderboard/leaderboard_resource.rs
anitarua: https://github.com/momentohq/client-sdk-java/blob/main/momento-sdk/src/main/java/momento/sdk/ILeaderboard.java#L92-L93

src/leaderboard/leaderboard_resource.rs
malandis: I will add that as for a follow up PR.

src/leaderboard/messages/data/length.rs
malandis: Yep -- comprehensive documentation is for follow-up work.

src/leaderboard/config/configuration.rs
malandis: I agree -- right now those are all shared. We can address in follow ups.

src/leaderboard/leaderboard_resource.rs
anitarua: This is just a docs nit though; initially misread, default rank would still be applied here

anitarua (approved):
lgtm!

rishtigupta:
lgtm too!
Will defer to other reviewers' approval though just because it is rust.

PR #439, opened 2025-02-11
feat: add competition ranking to leaderboard

virratanasangpunth:


src/leaderboard/messages/data/get_competition_rank.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl GetCompetitionRankResponse {
+    pub fn elements(&self) -> &[RankedElement] {
+        &self.elements
+    }
+
+    pub fn into_elements(self) -> Vec<RankedElement> {
+        self.elements
+    }
+}
```
malandis, line 64: No need for this since we can re-use `FetchResponse`

malandis (commented):
Couple changes upon review:

- An oversight from the previous PR: I should have re-used `FetchResponse` for `get_rank` instead of making a new response type. Could you refactor `get_competition_rank` to use `FetchResponse` instead of a new response type here as well?
- Also refactor `get_rank` to use the same

src/leaderboard/messages/data/delete.rs
```diff
@@ -27,7 +27,6 @@ impl LeaderboardRequest for DeleteRequest {
             cache_name,
             leaderboard.client_timeout(),
             momento_protos::leaderboard::DeleteLeaderboardRequest {
-                cache_name: cache_name.to_string(),
```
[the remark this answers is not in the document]
virratanasangpunth: yes that is correct.

src/leaderboard/leaderboard_resource.rs
```diff
@@ -103,6 +106,15 @@ impl Leaderboard {
         request.send(self).await
     }
 
+    pub async fn get_competition_rank(
+        &self,
+        ids: impl IntoIterator<Item = u32>,
+        rank: Option<Order>,
```
malandis, line 112: Could we remove the order here to be in line with `get_rank` (which doesn't take the order but can take it if the caller builds a `GetRankRequest` and calls `send`)? The rest of the PR is green from me.

src/leaderboard/messages/data/get_competition_rank.rs
```diff
@@ -10,12 +10,24 @@ pub struct GetCompetitionRankRequest {
...
     }
+
+    /// Sets the order ranking.
+    ///
+    /// Defaults to DESCENDING order.
+    pub fn order(mut self, order: Order) -> Self {
+        self.order = order;
+        self
+    }
 }
```
malandis, line 31: Should default to ascending following `get_rank` and the various fetch methods, eg in JS [see](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/internal/clients/leaderboard/AbstractLeaderboard.ts#L120-L132)

virratanasangpunth: Yea there was a discussion about this in the [PRD](https://www.notion.so/momentohq/PRD-Leaderboards-Competition-Ranking-17afeb15c97d8019aa15d508a8188083).
Our posture here is that for competition ranking so we should default to descending.

virratanasangpunth:
GH Actions seems to pick up older revisions.... odd....

virratanasangpunth:
https://github.com/momentohq/client-sdk-rust/actions/runs/13294066680/job/37121603883?pr=439

It's complaning about:
```
error[E0061]: this method takes 1 argument but 2 arguments were supplied
   --> tests/leaderboard/mod.rs:415:14
    |
415 |             .get_competition_rank([0, 1, 2, 3, 4], Some(Order::Ascending))
    |              ^^^^^^^^^^^^^^^^^^^^                  ---------------------- unexpected argument #2 of type `std::option::Option<momento::leaderboard::Order>`
```

but that doesn't match the code in this PR: https://github.com/momentohq/client-sdk-rust/pull/439/commits/5d07ce6c456093b2d5fbd66895b29f6307c9f7f4#diff-fd78100889d055010c44e3f81caf0225a1545b419395f83c7823dfd764eee13fR415

```
        let response = leaderboard
            .send_request(GetCompetitionRankRequest::new([0, 1, 2, 3, 4]).order(Order::Ascending))
            .await?;
```

src/leaderboard/messages/data/get_competition_rank.rs
malandis: Got it!

malandis (approved):

PR #456, opened 2025-04-30
feat: add sorted_set_union_store and sorted_set_length_by_score + upgrades

anitarua:
- **feat: add order optional argument to sorted_set_get_rank**
- **feat: add sorted_set_length_by_score**
- **fix sorted_set_length_by_score docs to specify optional arguments**
- **fix: add derive statement to CollectionTtl**
- **fix: upgrade client protos version**
- **feat: add sorted_set_union_store**


src/cache/cache_client.rs
```diff
@@ -1539,6 +1547,119 @@ impl CacheClient {
...
+    /// # }
+    /// ```
+    /// You can also use the [send_request](CacheClient::send_request) method to get an item using a [SortedSetLengthByScoreRequest]
+    /// which will allow you to set [optional arguments](SortedSetLengthByScoreRequest#optional-arguments) as well.
+    ///
+    /// For more examples of handling the response, see [SortedSetLengthByScoreResponse].
+    pub async fn sorted_set_length_by_score(
+        &self,
+        cache_name: impl Into<String>,
+        sorted_set_name: impl IntoBytes,
```
anitarua, line 1599: currently the required arguments are the same as for sorted_set_length

could include min_score and max_score as explicit Option arguments, meaning users would need to provide `None` if they don't want a max or min, but I suppose they wouldn't reach for this over sorted_set_length if they didn't intend to provide a score range as well?

malandis: I would just mirror the fetch by score API and request pattern.

anitarua: Cool, it's already matching then

malandis (approved):

PR #457, opened 2025-04-30
ci: make sure on-pull-request uses latest commit

anitarua:
Noticed on a recent PR that the second to last commit on the PR was being used for CI instead of the most recent commit. This PR will make it so the most recent commit on a PR is checked out. 

We should also make sure that the repo has enabled "Require branches to be up to date before merging".

FYI: we ran into this issue in the JS sdk as well and updated the workflow file in the same way ([reference](https://github.com/momentohq/client-sdk-javascript/pull/1506)).

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -34,7 +34,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          ref: ${{ github.event.pull_request.merge_commit_sha || github.event.pull_request.head.sha }}
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Log rust versions
         run: |
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -109,7 +109,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/on-pull-request.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          ref: ${{ github.event.pull_request.merge_commit_sha || github.event.pull_request.head.sha }}
+          ref: ${{ github.event.pull_request.head.sha }}
```
nand4011, line 37: Why is
```
with:
   ref: ${{ github.event.pull_request.head.sha }}
```
needed at all?

nand4011: Is it because were using
```
on:
  pull_request_target:
```
instead of
```
on:
  pull_request:
```
like I see in other repos?

anitarua: Yes, see this [doc](https://www.notion.so/momentohq/Approving-workflows-on-PRs-from-forks-175feb15c97d801d84eccc641b367a86?pvs=4#175feb15c97d808cadb8fda97c6f4d72) for more info

nand4011 (approved):
Good find

PR #458, opened 2025-05-02
fix: separate publish and subscribe channels in topic client

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1158

- Added a test to confirm topic client channel can become full and silently queue up requests
- Created separate pools of grpc channels in the topic client for publish and subscribe requests
- Bookkeeping for number of active subscriptions allows the topic client to return a limit exceeded error instead of silently queueing up requests

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -114,6 +114,19 @@ impl MomentoError {
             details: None,
         }
     }
+
+    pub(crate) fn max_concurrent_streams_reached(
+        num_grpc_channels: usize,
+        num_active_subscriptions: usize,
+        max_concurrent_streams: usize,
+    ) -> Self {
+        Self {
+            message: format!("Number of active streams: {}; number of grpc channels: {}; max concurrent streams: {}; Already at maximum number of concurrent grpc streams, cannot make new subscribe requests", num_active_subscriptions, num_grpc_channels, max_concurrent_streams),
+            error_code: MomentoErrorCode::LimitExceededError,
+            inner_error: None,
+            details: None,
+        }
+    }
 }
 
 /// Indicates an error source
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -122,7 +122,7 @@ impl MomentoError {
     ) -> Self {
         Self {
             message: format!("Number of active streams: {}; number of grpc channels: {}; max concurrent streams: {}; Already at maximum number of concurrent grpc streams, cannot make new subscribe requests", num_active_subscriptions, num_grpc_channels, max_concurrent_streams),
-            error_code: MomentoErrorCode::LimitExceededError,
+            error_code: MomentoErrorCode::ClientResourceExhausted,
             inner_error: None,
             details: None,
         }
--- a/src/topics/mod.rs
+++ b/src/topics/mod.rs
@@ -14,4 +14,3 @@ mod topic_client_builder;
 pub use topic_client::TopicClient;
 
 mod topic_subscription_manager;
-// pub use topic_subscription_manager::TopicSubscriptionManager;
--- a/src/topics/mod.rs
+++ b/src/topics/mod.rs
@@ -12,3 +12,6 @@ pub use config::configurations;
 mod topic_client;
 mod topic_client_builder;
 pub use topic_client::TopicClient;
+
+mod topic_subscription_manager;
+// pub use topic_subscription_manager::TopicSubscriptionManager;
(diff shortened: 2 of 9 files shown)


src/errors.rs
```diff
@@ -114,6 +114,19 @@ impl MomentoError {
...
     }
+
+    pub(crate) fn max_concurrent_streams_reached(
+        num_grpc_channels: usize,
+        num_active_subscriptions: usize,
+        max_concurrent_streams: usize,
+    ) -> Self {
+        Self {
+            message: format!("Number of active streams: {}; number of grpc channels: {}; max concurrent streams: {}; Already at maximum number of concurrent grpc streams, cannot make new subscribe requests", num_active_subscriptions, num_grpc_channels, max_concurrent_streams),
+            error_code: MomentoErrorCode::LimitExceededError,
```
malandis, line 125: Is this the right error for this? I thought we normally associate this with service limits as opposed to client limits.

src/topics/topic_client.rs
```diff
@@ -165,4 +172,79 @@ impl TopicClient {
...
+        if remaining_streams < 10 {
+            log::warn!(
+                "Only {} streams remaining.  You may hit the limit of {} concurrent streams soon.",
+                remaining_streams,
+                self.max_concurrent_streams
+            );
+        }
+
+        Ok(())
+    }
```
malandis, line 249: Is this our standard way of reporting this? What if a particular channel's streams are exhausted as opposed to the aggregate number of streams?

Ie if `any(count_streams(channel) for all clients) >= 100`?

src/errors.rs
anitarua: We went with LimitExceededError in the [go sdk](https://github.com/momentohq/client-sdk-go/blob/main/momento/pubsub_client.go#L113-L115). Could switch to using ClientResourceExhaustedError instead if that seems to make more sense?

src/topics/topic_client.rs
anitarua: "Standard" only in that this was the implementation first devised in the go sdk. But we can definitely change it here and in go if there's a better way since it's all internal.

The body of `get_next_streaming_client` (lines 195-211 of this file) should prevent any single channel from accumulating >100 streams, and in case of bursts of subscriptions, it allows for trying to acquire a stream on multiple channels. 
During bursts, I figured the atomic counter could go above 100 due to the way the atomic int is incremented before validating whether it can be returned. 
Thus `check_number_of_concurrent_streams` would return a false positive error if it detects a channel having atomic int >100 when there is actually enough capacity for the burst of streams once things settle down.

This was my reasoning at least, definitely open to a different approach to solving this!

src/topics/topic_client.rs
```diff
@@ -40,8 +43,12 @@ type ChannelType = InterceptedService<Channel, HeaderInterceptor>;
 /// ```
 #[derive(Clone, Debug)]
 pub struct TopicClient {
-    pub(crate) client: pubsub::pubsub_client::PubsubClient<ChannelType>,
+    pub(crate) unary_client_index: Arc<AtomicUsize>,
+    pub(crate) streaming_client_index: Arc<AtomicUsize>,
+    pub(crate) unary_clients: Vec<PubsubClient<InterceptedService<Channel, HeaderInterceptor>>>,
+    pub(crate) streaming_clients: Vec<TopicSubscriptionManager>,
     pub(crate) configuration: Configuration,
+    pub(crate) max_concurrent_streams: usize,
```
nand4011, line 51: Is there a situation in which this isn't the size of streaming clients * 100? Is it needed as a separate argument?

src/topics/mod.rs
```diff
@@ -12,3 +12,6 @@ pub use config::configurations;
 mod topic_client;
 mod topic_client_builder;
 pub use topic_client::TopicClient;
+
+mod topic_subscription_manager;
+// pub use topic_subscription_manager::TopicSubscriptionManager;
```
nand4011, line 17: Nit: probably don't need this.

src/errors.rs
malandis: That seems more appropriate.

src/topics/topic_client.rs
anitarua: fair, I can remove it from the struct

nand4011 (approved):

malandis (approved):

PR #463, opened 2025-05-13
feat: add options for exclusive min/max for sorted set fetch by score and length by score

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1230

Exposing support for exclusive min/max scores for `sorted_set_fetch_by_score` and `sorted_set_length_by_score`, which is needed for momento proxy.

Introduces a `ScoreBound` enum to specify inclusive or exclusive bounds.

### DIFF
--- a/src/cache/messages/data/sorted_set/sorted_set_length_by_score.rs
+++ b/src/cache/messages/data/sorted_set/sorted_set_length_by_score.rs
@@ -21,6 +21,9 @@ use crate::{
 ///   infinity.
 /// * `max_score` - the maximum score (inclusive) of the elements to fetch. Defaults to positive
 ///   infinity.
+/// * `exclusive_min_score` - the minimum score (exclusive) of the elements to fetch. Defaults to negative infinity.
+/// * `exclusive_max_score` - the maximum score (exclusive) of the elements to fetch. Defaults to positive
+///   infinity.
 ///
 /// # Examples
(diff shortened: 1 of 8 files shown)


malandis (commented):
The SDK is pre-1.0 so we can modify the API. I would advocate for either:

1. Change the builders to have `inclusive_min_score` and `exclusive_min_score` methods (same for max), or
2. Change the builders to have `min_score` and method to set the exclusive vs inclusive flag.

Doing this from scratch I prefer 1. But other SDKs don't expose this, so they will most likely look like 2.

src/cache/messages/data/sorted_set/sorted_set_length_by_score.rs
```diff
@@ -10,20 +10,28 @@ use crate::{
     MomentoResult,
 };
 
+/// Boundary for a sorted set score range.
+#[derive(Debug, PartialEq, Clone)]
+pub enum ScoreBound {
```
malandis, line 15: Not a blocker and can be done in a follow up. I'd put this in a separate module since it is a common concern of other requests.

malandis (dismissed):

src/cache/messages/data/sorted_set/sorted_set_length_by_score.rs
anitarua: can do it now while I'm here 👍

malandis (approved):

PR #467, opened 2025-07-02
chore: add Functions client

kvcache:
new client plumbing for the functions management apis

I'm doing one rpc first to make sure the pattern and client plumbing is right


PDXKimani (commented):
Seems alright to me, once we sort out build/test stuff.

src/functions/function_client.rs
```diff
@@ -0,0 +1,37 @@
...
+        &self,
+    ) -> &FunctionRegistryClient<InterceptedService<Channel, HeaderInterceptor>> {
+        &self.client
+    }
+
+    /// Send a Functions request
+    pub async fn send<R: MomentoRequest>(&self, request: R) -> MomentoResult<R::Response> {
+        request.send(self).await
+    }
+}
```
malandis, line 37: Our pattern is to provide a methods on the client for each RPC covering the most common request arguments. Developers wanting to customize the full suite of arguments can build a request and pass to `send`. So we should add a `put_function` method here.

src/functions/messages/put_function.rs
```diff
@@ -0,0 +1,130 @@
...
+            environment: HashMap::new(),
+            wasm_source: wasm_source.into(),
+        }
+    }
+
+    /// Set the Function's description
+    pub fn with_description(mut self, description: impl Into<String>) -> Self {
+        self.description = description.into();
+        self
+    }
```
malandis, line 64: Here's an example of [how we model](https://github.com/momentohq/client-sdk-rust/blob/main/src/cache/messages/data/sorted_set/sorted_set_fetch_by_score.rs#L66-L119) rich requests with defaults. Can apply that pattern here.

malandis (commented):
A couple comments re: existing patterns

src/functions/function.rs
```diff
@@ -0,0 +1,143 @@
...
+                }
+            }
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum WasmSource {
+    /// The wasm source is right here
+    Inline(Vec<u8>),
```
anitarua, line 105: also for my understanding: the wasm source is the binary that the Function will execute? and updating a Function's WasmSource would increment its version number?

src/functions/function_client.rs
kvcache: There is not as much of a "sensible default" for a Function as there is for the likes of `get(key)`. We can chat on slack, but I don't think we want to obscure configurations for Function definitions - even if it means that for consistency within the Functions client, all Requests are funneled through `send()`.

src/functions/function.rs
kvcache: Yes, it will increment the version number. The service manages sequence numbers per Cache.Function.

src/functions/function_client.rs
malandis: > There is not as much of a "sensible default" for a Function

That's not what I meant. Rather which fields for `PutFunctionRequest` are really mandatory vs what could one omit? Eg `cache_name`, `function_name`, `wasm_source` seems mandatory; `description` and `environment` optional. Thus there could be a method here the first four.

src/functions/messages/put_function.rs
malandis: Optional things are `Option`. That allows developers to pass in `None` (vs a default value) when instantiating the struct or using the setters. This was pattern was requested for ease of use with `rpc-perf`.

Setters are also named as just the name of field and accept an `Option`.

src/functions/function_client.rs
malandis: > Further, I regret my historical assent to the 2-way approach of the other apis. I'd propose that as part of the v1.0 exercise we should align everything to 1 pattern, and that pattern should be the request builders.

I'm cool with having One Way To Do It(tm). Having more than one way means people will either give up or not find the second more expressive way (in spite of rigorous documentation this comes up repeatedly).

> I am concerned about the loss of in-context educational documentation for Functions if we are to obscure other configurations than the 3 you mentioned.

I mentioned those three since the `new` on `PutFunctionRequest` had only those three, hence was making an opinion about the important ones. We can keep this as is for now.

No blocker and can always revisit/chat offline.

anitarua (approved):
I think having a `put_function(cache_name, function_name, wasm_source)` method on the function client would be nice, but not a blocker

PR #528, opened 2026-03-26
feat: Add list retain

nand4011:


malandis (commented):
LGTM so far but 1 question re: indices and a comment about the response data.

malandis (approved):

PR #530, opened 2026-04-15
feat: pin version with `PutFunctionConfigRequest`

kirmar:
This PR implements `PutFunctionConfigRequest`, which so far allows customers to pin their Momento Function to a specific version, instead of always following the latest version.

This feature is currently in beta/preview, alongside the other Functions features.

## Usage Examples

**Setup:**

1. Set your `cache_name`, `function_name`, `function_client`
2. Put your function

**Pin to a specific version:**

```rust
use momento::functions::{CurrentFunctionVersion, PutFunctionConfigRequest};

let request = PutFunctionConfigRequest::from_function_name(cache_name, function_name)
    .current_version(CurrentFunctionVersion::Pinned(0));
let function = function_client.send(request).await?;
```

**Pin to a specific version, concise style:**

```rust
use momento::functions::PutFunctionConfigRequest;

let request = PutFunctionConfigRequest::from_function_name(cache_name, function_name)
    .current_version(1));
let function = function_client.send(request).await?;
```

**Always use/follow the latest version:**

```rust
use momento::functions::{CurrentFunctionVersion, PutFunctionConfigRequest};

let request = PutFunctionConfigRequest::from_function_name(cache_name, function_name)
    .current_version(CurrentFunctionVersion::Latest);
let function = function_client.send(request).await?;
```

## Tests

**Setup:**

```bash
read -s -p "v2 API key: " MOMENTO_API_KEY
read -s -p "v1 API key: " V1_API_KEY
read -p "Endpoint: " MOMENTO_ENDPOINT

export MOMENTO_API_KEY V1_API_KEY MOMENTO_ENDPOINT
export TEST_CACHE_NAME=rust-sdk-test-cache TEST_AUTH_CACHE_NAME=rust-sdk-test-cache-auth
```

**First-time setup:**

```bash
momento cache create $TEST_CACHE_NAME
momento cache create $TEST_AUTH_CACHE_NAME
```

**Run doctests:**

```bash
cargo test --doc PutFunctionConfigRequest
# running 1 test
# test src/functions/messages/put_function_config.rs - functions::messages::put_function_config::PutFunctionConfigRequest (line 22) ... ok
# test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 209 filtered out; finished in 1.04s

make test-doctests
# test result: ok. 210 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 25.37s
```

### DIFF
--- a/src/functions/messages/put_function_config.rs
+++ b/src/functions/messages/put_function_config.rs
@@ -34,8 +34,7 @@ use momento_protos::function_types::FunctionKey;
 /// # let function = function_client.send(request).await?;
 /// # println!("Created a function: {function:?}");
 ///
-/// let request = PutFunctionConfigRequest::from_function_name(cache_name, "hello functions")
-///     .current_version(CurrentFunctionVersion::Pinned(0));
+/// let request = PutFunctionConfigRequest::from_function_name(cache_name, "hello functions").current_version(0);
 /// let function = function_client.send(request).await?;
 /// println!("Updated a function's config: {function:?}");
 /// # Ok(())
@@ -76,8 +75,11 @@ impl PutFunctionConfigRequest {
     }
 
     /// Choose the version to use upon invocation
-    pub fn current_version(mut self, current_version: CurrentFunctionVersion) -> Self {
-        self.new_version = Some(current_version.into());
+    pub fn current_version(mut self, current_version: impl Into<CurrentFunctionVersion>) -> Self {
+        let current_version: CurrentFunctionVersion = current_version.into();
+        let current_version: momento_protos::function_types::CurrentFunctionVersion =
+            current_version.into();
+        self.new_version = Some(current_version);
         self
     }
 }
--- a/src/functions/messages/put_function_config.rs
+++ b/src/functions/messages/put_function_config.rs
@@ -1,13 +1,13 @@
 use std::time::Duration;
 
 use crate::{
-    functions::{Function, FunctionClient, MomentoRequest},
+    functions::{CurrentFunctionVersion, Function, FunctionClient, MomentoRequest},
     utils::prep_request_with_timeout,
     MomentoError, MomentoResult,
 };
 
 use momento_protos::function::put_function_config_request::FunctionSpecifier;
-use momento_protos::function_types::{CurrentFunctionVersion, FunctionKey};
+use momento_protos::function_types::FunctionKey;
 
 /// Update a Function's configuration.
 /// The cache is used as a namespace for your Functions.
@@ -23,18 +23,13 @@ use momento_protos::function_types::{CurrentFunctionVersion, FunctionKey};
 /// # fn main() -> anyhow::Result<()> {
 /// # tokio_test::block_on(async {
 /// use momento::{CredentialProvider, FunctionClient};
-/// use momento::functions::PutFunctionConfigRequest;
-/// use momento_protos::function_types::{current_function_version, CurrentFunctionVersion};
+/// use momento::functions::{CurrentFunctionVersion, PutFunctionConfigRequest};
 /// # use momento_test_util::echo_wasm;
 /// # let (function_client, cache_name) = momento_test_util::create_doctest_function_client();
 /// // load your wasm from a .wasm file compiled with wasm32-wasip2
 /// let function_body = echo_wasm();
 ///
-/// let request = PutFunctionConfigRequest::new(cache_name, "hello functions").current_version(CurrentFunctionVersion {
-///     version: Some(current_function_version::Version::Pinned(current_function_version::Pinned {
-///         pinned_version: 0
-///     }))
-/// });
+/// let request = PutFunctionConfigRequest::new(cache_name, "hello functions").current_version(CurrentFunctionVersion::Pinned(0));
 /// let function = function_client.send(request).await?;
 /// println!("Updated a function's config: {function:?}");
 /// # Ok(())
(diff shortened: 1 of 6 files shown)


src/functions/messages/put_function_config.rs
```diff
@@ -0,0 +1,105 @@
...
+    function_specifier: FunctionSpecifier,
+    new_version: Option<momento_protos::function_types::CurrentFunctionVersion>,
+}
+
+impl PutFunctionConfigRequest {
+    /// Create a new PutFunctionConfigRequest, specified by function name
+    pub fn from_function_name(
+        cache_name: impl Into<String>,
+        function_name: impl Into<String>,
+    ) -> Self {
```
kirmar, line 50: I believe this is the first time we're allowing a customer to specify their Function however they prefer, either by name or ID. How should the interface look?

I had this earlier:
```rust
PutFunctionConfigRequest::new(cache_name).function_name("ping").current_version(CurrentVersion::Pinned(0))
```

I personally didn't like how that puts all the config fields (i.e. `current_version` so far) on the same "level" as the function being configured, so I changed it to this for now:
```rust
PutFunctionConfigRequest::from_function_name(cache_name, "ping").current_version(CurrentVersion::Pinned(0))
```

tylerburdsall: I quite like this design actually, it's direct and informs callers exactly the kind of key they are making. Nice touch!

tylerburdsall (dismissed):

src/functions/messages/put_function_config.rs
```diff
@@ -0,0 +1,111 @@
...
+use momento_protos::function::put_function_config_request::FunctionSpecifier;
+use momento_protos::function_types::FunctionKey;
+
+/// Update a Function's configuration.
+/// The cache is used as a namespace for your Functions.
+///
+/// # Arguments
+///
+/// * `cache_name` - The name of the cache to use as a namespace for the Function.
+/// * `function_name` - The name of the Function.
```
kvcache: Structs don't really have arguments, but the example looks good to me

src/functions/messages/put_function_config.rs
```diff
@@ -0,0 +1,111 @@
...
+/// # use momento_test_util::echo_wasm;
+/// # let (function_client, cache_name) = momento_test_util::create_doctest_function_client();
+/// # let function_body = echo_wasm();
+/// # let request = PutFunctionRequest::new(cache_name.clone(), "hello functions", function_body);
+/// # let function = function_client.send(request).await?;
+/// # println!("Created a function: {function:?}");
+///
+/// let request = PutFunctionConfigRequest::from_function_name(cache_name, "hello functions")
+///     .current_version(CurrentFunctionVersion::Pinned(0));
+/// let function = function_client.send(request).await?;
```
kvcache, line 33: I think this is going to fail at publish time. `docs.rs` does [sandboxed builds](https://docs.rs/releases/queue) for each release of every crate. I think you want `rust,no_run` on the doctest

src/functions/messages/put_function_config.rs
```diff
@@ -0,0 +1,111 @@
...
+        let cache_name = cache_name.into();
+        Self {
+            cache_name: cache_name.clone(),
+            function_specifier: FunctionSpecifier::FunctionId(function_id.into()),
+            new_version: None,
+        }
+    }
+
+    /// Choose the version to use upon invocation
+    pub fn current_version(mut self, current_version: CurrentFunctionVersion) -> Self {
```
kvcache: You could take `current_version: impl Into<CurrentFunctionVersion>` and `impl From<u32> for CurrentFunctionVersion` so people could write `builder.current_version(4)`

kvcache (commented):
looks good - a couple notes

src/functions/messages/put_function_config.rs
kirmar: Nice! Implemented in 74bfd7a

I couldn't directly convert `u32` to the proto `CurrentFunctionVersion` because "only traits defined in the current crate can be implemented for types defined outside of the crate," so I had to do a funny little `.into().into()`. Reorganized in 599ac8f, but let me know which style you prefer

src/functions/messages/put_function_config.rs
kirmar: Thanks! This comes directly from [put_function.rs](https://github.com/momentohq/client-sdk-rust/blob/8dc2c3a1b62a2dba53cbed4c75abb3c26f173711/src/functions/messages/put_function.rs#L12). Would you like me to update both files to use another word?

kvcache: IMO drop the arguments documentation in both places. It's just kind of a category error.

kvcache (dismissed):
Do as you like with the docs - change looks good. Thanks!

src/functions/messages/put_function_config.rs
kirmar: Ooh, interesting. Is this situation different from [put_function.rs](https://github.com/momentohq/client-sdk-rust/blob/8dc2c3a1b62a2dba53cbed4c75abb3c26f173711/src/functions/messages/put_function.rs#L19), [list_caches.rs](https://github.com/momentohq/client-sdk-rust/blob/8dc2c3a1b62a2dba53cbed4c75abb3c26f173711/src/cache/messages/control/list_caches.rs#L11), etc. because `PutFunctionConfigRequest`'s doctest uses `PutFunctionRequest` too? Or do you mean something else when you say the sandboxed builds may cause issues?

src/functions/messages/put_function_config.rs
kirmar: You know what, `grep` is giving me a total of 188 results :sweat_smile: so I shall save the other files for a separate cleanup PR. Can remove the list here though - also nice 'cause we don't have to remember to update it

tylerburdsall (approved):

PR #538, opened 2026-05-04
fix: `SubscribeRequest::sequence_page` copy-paste typo

kirmar:
`SubscriptionRequest::sequence_page` now takes the inner value of `SubscribeRequest::resume_at_sequence_page`, instead of using identical logic to  `resume_at_topic_sequence_number`.


anitarua (approved):
nice catch!

kvcache (approved):
thanks

PR #540, opened 2026-06-08
feat: offer eager connection establishment on CacheClient

anitarua:
Offer alternative build_and_connect() method to establish active grpc connections up front instead of lazy connections.

### DIFF
--- a/src/cache/messages/data/sorted_set/sorted_set_get_scores.rs
+++ b/src/cache/messages/data/sorted_set/sorted_set_get_scores.rs
@@ -125,8 +125,7 @@ impl<F: IntoBytesIterable + Clone> TryFrom<SortedSetGetScoresResponse<F>>
                 values, responses, ..
             } => {
                 let mut result = Vec::new();
-                for (value, response) in values.into_bytes().into_iter().zip(responses.into_iter())
-                {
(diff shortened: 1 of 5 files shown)


src/cache/messages/data/sorted_set/sorted_set_get_scores.rs
```diff
@@ -158,9 +157,7 @@ impl<F: IntoBytesIterable + Clone> TryFrom<SortedSetGetScoresResponse<F>>
                 values, responses, ..
             } => {
                 let mut result = Vec::new();
-                for (bytes_value, response) in
-                    values.into_bytes().into_iter().zip(responses.into_iter())
-                {
+                for (bytes_value, response) in values.into_bytes().into_iter().zip(responses) {
```
rlinehan, line 160: why were these two things changed? clippy updates?

rlinehan (approved):
I'm not very familiar with this code but this seems reasonable 🙈

src/cache/messages/data/sorted_set/sorted_set_get_scores.rs
anitarua: yeah the `.into_iter()` usage in these spots came up as lint errors

